PR #2, opened 2019-12-06
make minimal command worth trying

deads2k:
I've wired in bits to create a command that might run.


deads2k:
/hold

I need to save history and then clean up, not do it like this.

deads2k:
and here we go

PR #7, opened 2020-01-10
Fix startup

sttts:
- register flags
- cleanup cmd code
- add test server and startup test
- add meta/v1 types to apiserver schemes


pkg/cmd/auth-apiserver/cmd.go
```diff
@@ -3,14 +3,14 @@ package oauth_apiserver
 import (
 	"io"
 
+	"github.com/spf13/cobra"
```
p0lyn0mial: nit: add an empty space

pkg/cmd/auth-apiserver/cmd.go
```diff
@@ -39,22 +39,44 @@ func NewOAuthAPIServerOptions(out io.Writer) *OAuthAPIServerOptions {
 	}
 }
 
+func (o OAuthAPIServerOptions) Validate(args []string) error {
```
p0lyn0mial: nit: should be a pointer receiver?

pkg/cmd/auth-apiserver/cmd.go
sttts: How many import section do we want? 5? 6? :)

pkg/cmd/auth-apiserver/cmd.go
sttts: No, shouldn't. It's must not have side effects.

pkg/cmd/auth-apiserver/cmd.go
p0lyn0mial: we usually have:
```
std

other pkgs

kube

openshift
```

pkg/cmd/auth-apiserver/cmd.go
p0lyn0mial: interesting, what can cause side effects?

PR #9, opened 2020-03-30
Add token validation for OpenShift OAuth access tokens

stlaz:
Creates a new endpoint that can be used for authnenticating Kubernetes
users with webhook token authentication.

- [x] - bootstrap authenticator


pkg/apiserver/apiserver.go
```diff
@@ -2,6 +2,7 @@ package apiserver
 
 import (
 	"github.com/openshift/oauth-apiserver/pkg/serverscheme"
+	"github.com/openshift/oauth-apiserver/pkg/tokenvalidation"
```
vareti, line 5: split and sort imports?

pkg/cmd/oauth-apiserver/cmd.go
```diff
@@ -21,6 +21,7 @@ import (
 	// to force compiling
 	_ "github.com/openshift/oauth-apiserver/pkg/oauth/apiserver"
 	"github.com/openshift/oauth-apiserver/pkg/serverscheme"
+	tokenvalidationoptions "github.com/openshift/oauth-apiserver/pkg/tokenvalidation/options"
```
vareti, line 24: maybe group similar package imports?

pkg/tokenvalidation/validationconfig.go
```diff
@@ -0,0 +1,170 @@
...
+			minimumInactivityTimeoutSeconds)
+
+		validators = append(validators, timeoutValidator)
+
+		s.GenericAPIServer.AddPostStartHookOrDie("openshift.io-StartOAuthInformer", func(ctx genericapiserver.PostStartHookContext) error {
+			go oauthInformer.Start(ctx.StopCh)
+			return nil
+		})
+		s.GenericAPIServer.AddPostStartHookOrDie("openshift.io-StartTokenTimeoutUpdater", func(ctx genericapiserver.PostStartHookContext) error {
+			go oauthInformer.Start(ctx.StopCh)
```
vareti, line 147: is this intentional or you meant to start timeout validator?

stlaz: that was a mistake, it's fixed now

pkg/apiserver/apiserver.go
```diff
@@ -2,26 +2,30 @@ package apiserver
...
 type OAuthAPIServer struct {
 	GenericAPIServer *genericapiserver.GenericAPIServer
 }
 
 type completedConfig struct {
-	GenericConfig genericapiserver.CompletedConfig
-	ClientConfig  *restclient.Config
+	GenericConfig         genericapiserver.CompletedConfig
+	ClientConfig          *restclient.Config
+	TokenValidationConfig tokenvalidation.TokenValidationConfig
```
sttts, line 28: elsewhere we have the style of `ExtraConfig`, subsuming everything in addition. It makes it easier to just copy everything in the completion func.

stlaz:
/remove-lifecycle stale
this back in

pkg/tokenvalidation/timeoutvalidator.go
```diff
@@ -0,0 +1,233 @@
...
+	defaultTimeout time.Duration
+	tickerInterval time.Duration
+
+	// fields that are used to have a deterministic order of events in unit tests
+	flushHandler    func(flushHorizon time.Time) // allows us to decorate this func during unit tests
+	putTokenHandler func(td *tokenData)          // allows us to decorate this func during unit tests
+	clock           clock.Clock                  // allows us to control time during unit tests
+}
+
+func NewTimeoutValidator(tokens oauthclient.OAuthAccessTokenInterface, oauthClients oauthclientlister.OAuthClientLister, defaultTimeout int32, minValidTimeout int32) *TimeoutValidator {
```
p0lyn0mial: is there a hard limit after which a token will timeout?

pkg/tokenvalidation/tokenauthenticator.go
```diff
@@ -0,0 +1,84 @@
...
+	return &tokenAuthenticator{
+		tokens:       tokens,
+		users:        users,
+		groupMapper:  groupMapper,
+		validators:   OAuthTokenValidators(validators),
+		implicitAuds: implicitAuds,
+	}
+}
+
+func (a *tokenAuthenticator) AuthenticateToken(ctx context.Context, name string) (*kauthenticator.Response, bool, error) {
```
p0lyn0mial: ok, so this is a mapping between a `token` and a `user`.

pkg/tokenvalidation/tokenauthenticator.go
```diff
@@ -0,0 +1,84 @@
...
+	token, err := a.tokens.Get(context.TODO(), name, metav1.GetOptions{})
```
p0lyn0mial: an access token is set by `oauth-server` after successful login, correct ?

pkg/tokenvalidation/expirationvalidator.go
```diff
@@ -0,0 +1,31 @@
...
+	oauthv1 "github.com/openshift/api/oauth/v1"
+	userv1 "github.com/openshift/api/user/v1"
+)
+
+var errExpired = errors.New("token is expired")
+
+func NewExpirationValidator() OAuthTokenValidator {
+	return OAuthTokenValidatorFunc(
+		func(token *oauthv1.OAuthAccessToken, _ *userv1.User) error {
+			if token.ExpiresIn > 0 {
```
p0lyn0mial: is this a hard limit after which the token expires?

pkg/tokenvalidation/bootstrapauthenticator.go
```diff
@@ -0,0 +1,102 @@
...
+const ClusterAdminGroup = "system:cluster-admins"
+
+type bootstrapAuthenticator struct {
+	tokens            oauthclient.OAuthAccessTokenInterface
+	getter            bootstrap.BootstrapUserDataGetter
+	validator         OAuthTokenValidator
+	implicitAudiences kauthenticator.Audiences
+}
+
+func NewBootstrapAuthenticator(tokens oauthclient.OAuthAccessTokenInterface, getter bootstrap.BootstrapUserDataGetter, implicitAudiences kauthenticator.Audiences, validators ...OAuthTokenValidator) kauthenticator.Token {
```
p0lyn0mial: what the purpose of this authenticator?

stlaz: The purpose of this whole commit is to pull the authenticators from o/k to simplify review by showing how they changed. This specific authenticator gets merged into the main tokenvalidationhandler (https://github.com/openshift/oauth-apiserver/pull/9/commits/cfd5628e085a590e6d0eea0befdc94f327cf3605#diff-3e4e445b703e99b73d48bdb579ec3e69502289fa4245fef482ba16f91ab0d0eeR115) so that we finally stop making development mistakes by forgetting about this special cased authenticator.

Its original purpose was to authenticate the kube:admin user.

pkg/tokenvalidation/expirationvalidator.go
stlaz: Yes, the next two lines test it

pkg/tokenvalidation/timeoutvalidator.go
stlaz: last request time + whatever timeout is configured

pkg/tokenvalidation/tokenauthenticator.go
stlaz: the `oauth-server` creates it, yes

pkg/tokenvalidation/validationconfig.go
```diff
@@ -0,0 +1,165 @@
...
+	s.GenericAPIServer.AddPostStartHookOrDie("openshift.io-StartOAuthInformer", func(ctx genericapiserver.PostStartHookContext) error {
+		go oauthInformer.Start(ctx.StopCh)
+		return nil
+	})
+	s.GenericAPIServer.AddPostStartHookOrDie("openshift.io-StartTokenTimeoutUpdater", func(ctx genericapiserver.PostStartHookContext) error {
+		go timeoutValidator.Run(ctx.StopCh)
+		return nil
+	})
+
+	validationHandler := NewTokenValidationHandler(
```
deads2k, line 146: I suggest adding a strategy like we did for the TokenReviews upstream.  Doing it that way avoids adding all this and simply adds a new resource to where you're building your existing one.

You could have

`/apis/oauth.openshift.io/v1/oauthtokenreviews/`

deads2k: See https://github.com/kubernetes/kubernetes/blob/07d88617e55b64c243956c7b2032430beb03b159/pkg/registry/authentication/tokenreview/storage.go for example.

deads2k:
I'd like to see it implemented like TokenReviews upstream.

stlaz:
I was not able to find a workaround around all the generic API serving protections so that I could serve k8s tokenreview code

pkg/apiserver/apiserver.go
```diff
@@ -57,6 +68,11 @@ func (c completedConfig) New(delegationTarget genericapiserver.DelegationTarget)
 	delegateAPIServer := delegationTarget
 	var err error
 
+	delegateAPIServer, err = c.withTokenValidation(delegateAPIServer)
```
sttts, line 71: TokenReview or TokenValidation ?

pkg/tokenvalidation/validationconfig.go
stlaz: @deads2k  I haven't looked much into generating the openapi models needed in order to be able to directly serve the actual k8s.TokenReview object needed by the rest endpoint, but the other way around (mirroring the TokenReview object in the openshift/api) proved unfruitful.

I used the type from my https://github.com/stlaz/api/commit/584f090a52af86aac4b61067ab7730c446f031c8, however, after wiring it similarly to what upstream does (basically just replacing the type in the `REST.New()` method), the generic handlers eventually won't pass the object to `REST.Create()` because the `gvk` of the object sent by kube does not match (the groups differ) - [kube-apiserver crete handler gvk check](https://github.com/kubernetes/apiserver/blob/0e46f0ea2bdc958827712205cf9a88140a28409d/pkg/endpoints/handlers/create.go#L125).

That leads me to believe that some kind of a prelude handler might always be necessary. I could hack around the generic API machinery handlers by keeping the below code but instead of `validationHandler` have a wrapper that would call the upstream's `tokenreview.REST.Create()` with the object from the incoming request. If that's what you'd prefer.

Or maybe you had something simpler in mind, my solutions may oftentimes be too complicated simply because I don't know what to look for.

stlaz:
added an implementation that wraps the upstream REST storage object to workaround for the generic serving code validations

hack/openapi-violation.list
```diff
@@ -1,3 +1,25 @@
...
+API rule violation: list_type_missing,github.com/openshift/api/authorization/v1,Role,Rules
+API rule violation: list_type_missing,github.com/openshift/api/authorization/v1,RoleBinding,Subjects
+API rule violation: list_type_missing,github.com/openshift/api/authorization/v1,ServiceAccountRestriction,Namespaces
+API rule violation: list_type_missing,github.com/openshift/api/authorization/v1,ServiceAccountRestriction,ServiceAccounts
+API rule violation: list_type_missing,github.com/openshift/api/authorization/v1,SubjectAccessReview,GroupsSlice
+API rule violation: list_type_missing,github.com/openshift/api/authorization/v1,SubjectRulesReviewSpec,Groups
+API rule violation: list_type_missing,github.com/openshift/api/authorization/v1,SubjectRulesReviewStatus,Rules
+API rule violation: list_type_missing,github.com/openshift/api/authorization/v1,UserRestriction,Groups
+API rule violation: list_type_missing,github.com/openshift/api/authorization/v1,UserRestriction,Selectors
+API rule violation: list_type_missing,github.com/openshift/api/authorization/v1,UserRestriction,Users
```
sttts, line 22: where do these come from?

pkg/oauth/apiserver/apiserver.go
```diff
@@ -3,32 +3,52 @@ package apiserver
...
 type ExtraConfig struct {
 	ServiceAccountMethod string
 
-	makeV1Storage sync.Once
-	v1Storage     map[string]rest.Storage
-	v1StorageErr  error
+	makeV1Storage      sync.Once
+	v1Storage          map[string]rest.Storage
+	v1StorageErr       error
+	postStartHooks     map[string]genericapiserver.PostStartHookFunc
```
sttts, line 48: new line above the hook.

pkg/oauth/apiserver/apiserver.go
```diff
@@ -72,7 +92,7 @@ func (c completedConfig) New(delegationTarget genericapiserver.DelegationTarget)
 		GenericAPIServer: genericServer,
 	}
 
-	v1Storage, err := c.V1RESTStorage()
+	v1Storage, tokenReviewService, postStartHooks, err := c.V1RESTStorage()
```
sttts, line 95: isn't tokenreview also a v1Storage?

pkg/oauth/apiserver/apiserver.go
```diff
@@ -83,21 +103,88 @@ func (c completedConfig) New(delegationTarget genericapiserver.DelegationTarget)
...
-		c.ExtraConfig.v1Storage, c.ExtraConfig.v1StorageErr = c.newV1RESTStorage()
+		c.ExtraConfig.v1Storage, c.ExtraConfig.tokenReviewService, c.ExtraConfig.postStartHooks, c.ExtraConfig.v1StorageErr = c.newV1RESTStorage()
 	})
 
-	return c.ExtraConfig.v1Storage, c.ExtraConfig.v1StorageErr
+	return c.ExtraConfig.v1Storage, c.ExtraConfig.tokenReviewService, c.ExtraConfig.postStartHooks, c.ExtraConfig.v1StorageErr
 }
 
-func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
+func (c *completedConfig) tokenReviewStorage(
```
sttts, line 122: am confused. Wasn't this supposed to be replaced by the "real" api?

pkg/oauth/apiserver/registry/oauthtokenreview/rest/tokenreviewwrapper.go
```diff
@@ -0,0 +1,94 @@
...
+	wrapped *ktokenreview.REST
+}
+
+func NewTokenReviewRESTWrapper(authenticator authenticator.Request) *TokenReviewRESTWrapper {
+	return &TokenReviewRESTWrapper{
+		wrapped: ktokenreview.NewREST(authenticator, []string{}),
+	}
+}
+
+func (tw *TokenReviewRESTWrapper) ServeHTTP(r *restful.Request, w *restful.Response) {
```
sttts, line 46: this looks differently to other rest implementations. Would expect something like in k/k/pkg/registry/authz/subjectaccessreview.

stlaz:
/test e2e-component
install failed on infra

stlaz:
fixed unit tests for the tokenvalidator

pkg/tokenreviews/apiserver.go
```diff
@@ -0,0 +1,187 @@
...
+	genericServer, err := c.GenericConfig.New("tokenreviews.openshift.io-apiserver", delegationTarget)
+	if err != nil {
+		return nil, err
+	}
+
+	s := &TokenReviewsAPIServer{
+		GenericAPIServer: genericServer,
+	}
+
+	v1Storage, postStartHooks, err := c.V1RESTStorage()
```
deads2k, line 87: We want this part of the storage added at `apis/oauth.openshift.io/v1/tokenaccessreviews`, but accepting the kube-apiserver format.  

See https://github.com/openshift/generic-admission-server/blob/master/pkg/registry/admissionreview/admission_review.go#L33-L45 for how to accept one version of a think in REST handler and serve from a different spot.  Very close now.

pkg/tokenreviews/apiserver.go
```diff
@@ -0,0 +1,187 @@
...
+		go oauthInformer.Start(ctx.StopCh)
+		return nil
+	}
+	c.ExtraConfig.postStartHooks["openshift.io-StartTokenTimeoutUpdater"] = func(ctx genericapiserver.PostStartHookContext) error {
+		go timeoutValidator.Run(ctx.StopCh)
+		return nil
+	}
+
+	tokenAuthenticator := tokenvalidation.NewTokenValidationHandler(
+		oauthClient.OauthV1().OAuthAccessTokens(), bootstrapUserDataGetter, userClient.UserV1().Users(), groupMapper,
```
deads2k, line 155: is this congruent to what we have in o/k today?  I'm surprised we use a real client to get oauth tokens.

pkg/tokenreviews/tokenvalidation/validators/timeoutvalidator_test.go
```diff
@@ -171,27 +103,25 @@ func TestAuthenticateTokenTimeout(t *testing.T) {
...
 	go timeouts.Run(stopCh)
 
 	// TIME: 0 seconds have passed here
 
 	// first time should succeed for all
-	checkToken(t, "testToken", tokenAuthenticator, accessTokenGetter, testClock, true)
+	checkToken(t, "testToken", timeouts, accessTokenGetter, testClock, false)
 	wait(t, putTokenSync)
 
-	checkToken(t, "quickToken", tokenAuthenticator, accessTokenGetter, testClock, true)
```
deads2k: why are these changing?  I'd rather try to limit changes in this first stage.

deads2k:
This needs an e2e test or integration test or something to show that you're serving this new type.  If stefan can live with it coming after, so could I.

Please prefix any commit with straight moves as `MOVE`.  It's a little difficult to determine full consistency, but this is looking pretty good.

pkg/tokenreviews/apiserver.go
stlaz: yes, apparently we're still doing live lookups on authentication request:
https://github.com/openshift/kubernetes/blob/4b40bb4792d0b183fd02636660c55faac08a8735/pkg/kubeapiserver/authenticator/patch_authenticator.go#L79

pkg/tokenreviews/tokenvalidation/validators/timeoutvalidator_test.go
stlaz: The test was wrong, rather than checking the TimeoutValidator, it was checking whether the whole TokenAuthenticator worked.

I separated validators into their own package so that they form their own logical unit, and to underline that they are pluggable to the authenticator. If I kept the original behavior of the test, this would have been a cyclic import.

Instead, I added a bunch of new unit tests for the whole authenticator in the above package (see `TestTokenAuthenticator`)

stlaz:
The comments should be addressed, I even added an e2e test that checks the apiserver's tokenreviews endpoint.

stlaz:
/retest
irrelevant

pkg/tokenvalidation/bootstrapauthenticator.go
```diff
@@ -0,0 +1,102 @@
...
+	kauthenticator "k8s.io/apiserver/pkg/authentication/authenticator"
+	kuser "k8s.io/apiserver/pkg/authentication/user"
+
+	authorizationv1 "github.com/openshift/api/authorization/v1"
+	userv1 "github.com/openshift/api/user/v1"
+	oauthclient "github.com/openshift/client-go/oauth/clientset/versioned/typed/oauth/v1"
+	bootstrap "github.com/openshift/library-go/pkg/authentication/bootstrapauthenticator"
+)
+
+const ClusterAdminGroup = "system:cluster-admins"
```
sttts: does it have to be public?

pkg/tokenvalidation/bootstrapauthenticator.go
```diff
@@ -0,0 +1,102 @@
...
+	return &bootstrapAuthenticator{
+		tokens:            tokens,
+		getter:            getter,
+		validator:         OAuthTokenValidators(validators),
+		implicitAudiences: implicitAudiences,
+	}
+}
+
+func (a *bootstrapAuthenticator) AuthenticateToken(ctx context.Context, name string) (*kauthenticator.Response, bool, error) {
+	token, err := a.tokens.Get(context.TODO(), name, metav1.GetOptions{})
```
sttts: why not the given ctx?

pkg/tokenvalidation/bootstrapauthenticator.go
sttts: disregard if this is fixed later-on. Still in the first commit.

pkg/apiserver/apiserver.go
```diff
@@ -17,6 +19,11 @@ import (
 
 type Config struct {
 	GenericConfig *genericapiserver.RecommendedConfig
+	ExtraConfig   OAuthAPIExtraConfig
+}
+
+type OAuthAPIExtraConfig struct {
+	AccessTokenInactivityTimeout time.Duration
```
sttts, line 26: comments

pkg/oauth/apiserver/apiserver.go
```diff
@@ -3,24 +3,39 @@ package apiserver
...
 	useroauthaccesstokensdelegate "github.com/openshift/oauth-apiserver/pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate"
 	"github.com/openshift/oauth-apiserver/pkg/serverscheme"
+	"github.com/openshift/oauth-apiserver/pkg/tokenreviews/tokenvalidation"
+	"github.com/openshift/oauth-apiserver/pkg/tokenreviews/tokenvalidation/usercache"
+	tokenvalidators "github.com/openshift/oauth-apiserver/pkg/tokenreviews/tokenvalidation/validators"
+)
+
+const (
+	defaultInformerResyncPeriod     = 10 * time.Minute
+	minimumInactivityTimeoutSeconds = 300
```
sttts, line 38: these are the old values from kube-apiserver I assume.

pkg/oauth/apiserver/apiserver.go
```diff
@@ -29,7 +44,12 @@ type ExtraConfig struct {
 	makeV1Storage sync.Once
 	v1Storage     map[string]rest.Storage
 	v1StorageErr  error
+
+	postStartHooks map[string]genericapiserver.PostStartHookFunc
```
sttts, line 48: don't we have these in the generic apiserver already? Why here too?

pkg/tokenvalidation/bootstrapauthenticator.go
stlaz: It gets removed eventually. The first commit is just a clean move.

pkg/tokenvalidation/bootstrapauthenticator.go
stlaz: Somebody probably overlooked this while wiring contexts in the 1.18 kube rebase. It gets removed in a later commit anyway.

pkg/oauth/apiserver/apiserver.go
stlaz: Yes:
https://github.com/openshift/kubernetes/blob/4b40bb4792d0b183fd02636660c55faac08a8735/cmd/kube-controller-manager/app/patch_informers_openshift.go#L171
https://github.com/openshift/kubernetes/blob/4b40bb4792d0b183fd02636660c55faac08a8735/pkg/kubeapiserver/authenticator/patch_authenticator.go#L65

pkg/oauth/apiserver/apiserver.go
stlaz: there's a kind of weird do-once logic in 
`func (c *completedConfig) V1RESTStorage()`
that I did not want to touch

pkg/oauth/apiserver/apiserver.go
```diff
@@ -83,21 +103,25 @@ func (c completedConfig) New(delegationTarget genericapiserver.DelegationTarget)
...
 
-func (c *completedConfig) V1RESTStorage() (map[string]rest.Storage, error) {
+func (c *completedConfig) V1RESTStorage() (map[string]rest.Storage, map[string]genericapiserver.PostStartHookFunc, error) {
 	c.ExtraConfig.makeV1Storage.Do(func() {
-		c.ExtraConfig.v1Storage, c.ExtraConfig.v1StorageErr = c.newV1RESTStorage()
+		c.ExtraConfig.v1Storage, c.ExtraConfig.postStartHooks, c.ExtraConfig.v1StorageErr = c.newV1RESTStorage()
 	})
 
-	return c.ExtraConfig.v1Storage, c.ExtraConfig.v1StorageErr
+	return c.ExtraConfig.v1Storage, c.ExtraConfig.postStartHooks, c.ExtraConfig.v1StorageErr
```
sttts, line 118: this func is the reason that you have the hooks in ExtraConfig?

sttts: Mutating `c.ExtraConfig.v1Storage` (preexisting) is already strange.

sttts: Can you clean that up? This is not right. And if we have a good reason for this unexpected state, there should be a clear comment.

pkg/oauth/apiserver/apiserver.go
```diff
@@ -107,49 +131,109 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+		oauthClient.OauthV1().OAuthAccessTokens(),
+		oauthInformer.Oauth().V1().OAuthClients().Lister(),
+		c.ExtraConfig.AccessTokenInactivityTimeout,
+		minimumInactivityTimeoutSeconds)
+
+	if c.ExtraConfig.postStartHooks == nil {
+		c.ExtraConfig.postStartHooks = map[string]genericapiserver.PostStartHookFunc{}
+	}
+
+	c.ExtraConfig.postStartHooks["openshift.io-StartUserInformer"] = func(ctx genericapiserver.PostStartHookContext) error {
```
sttts, line 219: return these, don't put them into the config

pkg/tokenreviews/tokenvalidation/tokenauthenticator.go
```diff
@@ -0,0 +1,191 @@
...
+	authenticatedOAuthGroup = "system:authenticated:oauth"
+
+	sha256Prefix = "sha256~"
+)
+
+var errLookup = errors.New("token lookup failed")
+
+var _ authenticator.Request = &TokenAuthenticator{}
+
+type TokenAuthenticator struct {
```
sttts, line 39: this file is 1:1 copy? Please put in extra commit.

pkg/oauth/apiserver/apiserver.go
stlaz: I removed those lines here.

pkg/oauth/apiserver/registry/tokenreviews/etcd.go
```diff
@@ -0,0 +1,55 @@
...
+	kauthenticationv1 "k8s.io/api/authentication/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/authentication/authenticator"
+	"k8s.io/apiserver/pkg/registry/rest"
+	kauthinternal "k8s.io/kubernetes/pkg/apis/authentication"
+	kauthv1internal "k8s.io/kubernetes/pkg/apis/authentication/v1"
+	"k8s.io/kubernetes/pkg/registry/authentication/tokenreview"
```
sttts, line 16: is this dependency new?

pkg/oauth/apiserver/registry/tokenreviews/etcd.go
```diff
@@ -0,0 +1,55 @@
...
+	if !ok {
+		return nil, apierrors.NewBadRequest(fmt.Sprintf("not a TokenReview: %#v", obj))
+	}
+
+	// convert to internal types here to avoid "cannot find model definition for %v" errors when using internal types in New()
+	tokenReviewInternal := &kauthinternal.TokenReview{}
+	if err := kauthv1internal.Convert_v1_TokenReview_To_authentication_TokenReview(tokenReview, tokenReviewInternal, nil); err != nil {
+		return nil, apierrors.NewInternalError(fmt.Errorf("failed to convert %#v to internal TokenReview: %v", obj, err))
+	}
+	return r.wrapped.Create(ctx, tokenReviewInternal, validateObj, createOptions)
```
sttts, line 54: this confuses me. Is this calling the upstream implementation or do we only use its interface?

pkg/tokenreviews/tokenvalidation/tokenauthenticator.go
```diff
@@ -1,84 +1,191 @@
...
+	if len(authorizationHeader) == 0 {
+		return nil, false, nil
 	}
 
-	user, err := a.users.Get(context.TODO(), token.UserName, metav1.GetOptions{})
-	if err != nil {
-		return nil, false, err
+	authzHeaderSplit := strings.Split(authorizationHeader, " ")
+	if len(authzHeaderSplit) < 2 || strings.ToLower(authzHeaderSplit[0]) != "bearer" {
+		return nil, false, nil
```
sttts, line 73: why all these changes here? I thought it's a direct copy from apiserver-library-go?!

pkg/tokenreviews/tokenvalidation/validators/timeoutvalidator_test.go
```diff
@@ -0,0 +1,272 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/clock"
+
+	oauthv1 "github.com/openshift/api/oauth/v1"
+	userv1 "github.com/openshift/api/user/v1"
+	oauthfake "github.com/openshift/client-go/oauth/clientset/versioned/fake"
+	oauthclient "github.com/openshift/client-go/oauth/clientset/versioned/typed/oauth/v1"
+)
+
+func TestAuthenticateTokenTimeout(t *testing.T) {
```
sttts, line 18: aren't there also just copied from somewhere?

go.mod
```diff
@@ -26,6 +27,7 @@ require (
 	k8s.io/client-go v0.20.1
 	k8s.io/code-generator v0.20.1
 	k8s.io/component-base v0.20.1
+	k8s.io/klog/v2 v2.4.0
 	k8s.io/kube-aggregator v0.20.1
 	k8s.io/kube-openapi v0.0.0-20201113171705-d219536bb9fd
 	k8s.io/kubernetes v1.20.1
```
sttts, line 33: for what did we need this dependency in the past?

pkg/oauth/apiserver/registry/tokenreviews/etcd.go
stlaz: yes, it bears the upstream implementation to handle the endpoint

pkg/oauth/apiserver/registry/tokenreviews/etcd.go
stlaz: it's calling the upstream implementation

pkg/tokenreviews/tokenvalidation/tokenauthenticator.go
stlaz: the authenticator for the upstream tokenreview registry has a slightly different interface (`"k8s.io/apiserver/pkg/authentication/authenticator"`) with slightly different expectations to the authenticator we had before (`"k8s.io/apiserver/pkg/authentication/authenticator"`).

The changes that follow only accommodate and leverage this change by unifying the two OpenShift-specific token authenticators we used to have into a single user-retrieving mechanism, duality of which has caused us such headaches in the past.

pkg/tokenreviews/tokenvalidation/validators/timeoutvalidator_test.go
stlaz: Those needed to be split and fixed so that they each tested the unit they were supposed to test: https://github.com/openshift/oauth-apiserver/pull/9#discussion_r568620634

go.mod
stlaz: k8s.io/kubernetes? I am not sure :/

pkg/oauth/apiserver/apiserver.go
```diff
@@ -95,49 +115,106 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+
+func (c *completedConfig) tokenReviewStorage(
+	corev1Client corev1.CoreV1Interface,
+	oauthClient *oauthclients.Clientset,
+	userClient *userclient.Clientset,
+) (rest.Storage, map[string]genericapiserver.PostStartHookFunc, error) {
+	bootstrapUserDataGetter := bootstrap.NewBootstrapUserDataGetter(corev1Client, corev1Client)
+
+	// create informer for the users to be used in user <-> groups mapping
+	userInformer := userinformer.NewSharedInformerFactory(userClient, defaultInformerResyncPeriod)
```
deads2k, line 183: Add these informerfactories  to your extraconfig object so that in the future you won't accidentally create another copy

pkg/oauth/apiserver/apiserver.go
```diff
@@ -95,49 +115,106 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+	if err := userInformer.User().V1().Groups().Informer().AddIndexers(cache.Indexers{
+		usercache.ByUserIndexName: usercache.ByUserIndexKeys,
+	}); err != nil {
+		return nil, nil, err
+	}
+
+	groupMapper := usercache.NewGroupCache(userInformer.User().V1().Groups())
+	oauthInformer := oauthinformer.NewSharedInformerFactory(oauthClient, defaultInformerResyncPeriod)
```
deads2k, line 191: same with this informer factory.

pkg/oauth/apiserver/registry/tokenreviews/etcd.go
deads2k: Copy/paste instead please.  The only k/k deps we have are on legacy scheme and life becomes easier if we split I think.

stlaz: Are you sure? This will cut the dependency so if there's a bug in this potentially security sensitive code we won't know, or if we're gone by that time noone else might realize they need to fix it.

IIUC copy/pasting would imply copying all of
"k8s.io/kubernetes/pkg/apis/authentication"
"k8s.io/kubernetes/pkg/registry/authentication/tokenreview"

go.mod
stlaz: https://github.com/openshift/oauth-apiserver/pull/9#discussion_r574518257

pkg/tokenreviews/tokenvalidation/validators/expirationvalidator_test.go
```diff
@@ -1,72 +1,50 @@
...
 		if err != errExpired {
 			t.Errorf("Unexpected error: %v", err)
 		}
-		if userInfo != nil {
-			t.Errorf("Unexpected user: %v", userInfo)
-		}
 	}
 }
 
-func TestAuthenticateTokenValidated(t *testing.T) {
```
deads2k: I don't see where this moved to

pkg/tokenreviews/tokenvalidation/validators/expirationvalidator_test.go
```diff
@@ -1,72 +1,50 @@
...
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	oauthv1 "github.com/openshift/api/oauth/v1"
 	userv1 "github.com/openshift/api/user/v1"
-	oauthfake "github.com/openshift/client-go/oauth/clientset/versioned/fake"
-	userfake "github.com/openshift/client-go/user/clientset/versioned/fake"
 )
 
-func TestAuthenticateTokenExpired(t *testing.T) {
```
deads2k: or this one

pkg/oauth/apiserver/apiserver.go
```diff
@@ -95,49 +144,103 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+func (c *completedConfig) tokenReviewStorage(
+	corev1Client corev1.CoreV1Interface,
+	oauthClient *oauthclients.Clientset,
+	userClient *userclient.Clientset,
+) (rest.Storage, map[string]genericapiserver.PostStartHookFunc, error) {
+	bootstrapUserDataGetter := bootstrap.NewBootstrapUserDataGetter(corev1Client, corev1Client)
+
+	// add indexes to the userinformer for the users to be used in user <-> groups mapping
+	userInformer := c.ExtraConfig.UserInformers
+	if err := userInformer.User().V1().Groups().Informer().AddIndexers(cache.Indexers{
```
deads2k, line 201: move this up where you initialize the informers.  this extra index is pretty integral.

pkg/oauth/apiserver/apiserver.go
```diff
@@ -95,49 +144,103 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+	oauthInformer := c.ExtraConfig.OAuthInformers
+
+	timeoutValidator := tokenvalidators.NewTimeoutValidator(
+		oauthClient.OauthV1().OAuthAccessTokens(),
+		oauthInformer.Oauth().V1().OAuthClients().Lister(),
+		c.ExtraConfig.AccessTokenInactivityTimeout,
+		minimumInactivityTimeoutSeconds)
+
+	postStartHooks := map[string]genericapiserver.PostStartHookFunc{}
+	postStartHooks["openshift.io-StartUserInformer"] = func(ctx genericapiserver.PostStartHookContext) error {
```
deads2k, line 217: this poststart hook can move to where you create the informer

pkg/oauth/apiserver/apiserver.go
```diff
@@ -95,49 +144,103 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+		go userInformer.Start(ctx.StopCh)
+		return nil
+	}
+	postStartHooks["openshift.io-StartOAuthInformer"] = func(ctx genericapiserver.PostStartHookContext) error {
```
deads2k, line 221: this poststart hook can move to where you create the informer

pkg/oauth/apiserver/apiserver.go
```diff
@@ -95,49 +144,103 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+		go oauthInformer.Start(ctx.StopCh)
+		return nil
+	}
+	postStartHooks["openshift.io-StartTokenTimeoutUpdater"] = func(ctx genericapiserver.PostStartHookContext) error {
```
deads2k, line 225: leave this one here.

pkg/oauth/apiserver/apiserver.go
```diff
@@ -95,49 +144,103 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+		go timeoutValidator.Run(ctx.StopCh)
+		return nil
+	}
+
+	// create the authenticator for the "normal" OpenShift OAuth2 users
+	tokenAuthenticator := tokenvalidation.NewTokenAuthenticator(
```
deads2k, line 231: did you combine `k8s.io/kubernetes/openshift-kube-apiserver/authentication/oauth` with something?

pkg/oauth/apiserver/apiserver.go
```diff
@@ -95,49 +144,103 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+		oauthClient.OauthV1().OAuthAccessTokens(), userClient.UserV1().Users(), groupMapper, c.ExtraConfig.APIAudiences,
+		tokenvalidators.NewExpirationValidator(), tokenvalidators.NewUIDValidator(), timeoutValidator)
```
deads2k, line 233: the previous `validators := []oauth.OAuthTokenValidator{oauth.NewExpirationValidator(), oauth.NewUIDValidator(), timeoutValidator}` was easier to read

pkg/oauth/apiserver/apiserver.go
deads2k, line 231: this is the `oauthTokenAuthenticator` from https://github.com/openshift/kubernetes/blob/master/pkg/kubeapiserver/authenticator/patch_authenticator.go#L67, right?

pkg/oauth/apiserver/apiserver.go
```diff
@@ -95,49 +144,103 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+		"oAuthAccessTokens":         accessTokenStorage,
+		"oAuthClients":              clientStorage,
+		"oAuthClientAuthorizations": clientAuthorizationStorage,
+		"userOAuthAccessTokens":     userOAuthAccessTokensDelegate,
+		"tokenReviews":              tokenReviewStorage,
+	}
+	return v1Storage, tokenReviewPostStartHooks, nil
+}
+
+func (c *completedConfig) tokenReviewStorage(
```
deads2k, line 192: @sttts this is the meat.  It needs to be perfectly congruent to https://github.com/openshift/kubernetes/blob/master/pkg/kubeapiserver/authenticator/patch_authenticator.go#L29

pkg/oauth/apiserver/apiserver.go
stlaz: it's more like the openshift authenticator from the patch we were adding combined with what's in "k8s.io/kubernetes/pkg/kubeapiserver/authenticator/config.go"

stlaz: yes

stlaz:
@deads2k @sttts I added `func (c *completedConfig) getOpenShiftAuthenticators()` to explicitly show which part comes from https://github.com/openshift/kubernetes/blob/master/pkg/kubeapiserver/authenticator/patch_authenticator.go#L29. I basically copy-pasted that and went line by line. This should now hopefully be easy to review.

I later combine the authenticator according to how upstream does it in `k8s.io/kubernetes/pkg/kubeapiserver/authenticator/config.go` **minus**:
- the token cache - we already get cached on KAS twice - tokens cache and webhooktokenauthenticator cache; I don't expect anything outside KAS directly reaching oauth-apiserver endpoint
- websocket authenticator - seemed unnecessary
- `system:authenticated` group adder - KAS will do it itself
- anonymous authenticator

pkg/oauth/apiserver/apiserver.go
```diff
@@ -95,49 +163,101 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+func (c *completedConfig) getOpenShiftAuthenticators(
+	corev1Client corev1.CoreV1Interface,
+	oauthClient *oauthclients.Clientset,
+	userClient *userclient.Clientset,
+) ([]authenticator.Token, map[string]genericapiserver.PostStartHookFunc) {
+	tokenAuthenticators := []authenticator.Token{}
+
+	bootstrapUserDataGetter := bootstrap.NewBootstrapUserDataGetter(corev1Client, corev1Client)
+
+	oauthInformer := oauthinformer.NewSharedInformerFactory(oauthClient, defaultInformerResyncPeriod)
```
deads2k, line 233: oauthInformer := c.ExtraConfig.OAuthInformers

pkg/oauth/apiserver/apiserver.go
```diff
@@ -95,49 +163,101 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+	userInformer := userinformer.NewSharedInformerFactory(userClient, defaultInformerResyncPeriod)
```
deads2k, line 234: userInformer := c.ExtraConfig.UserInformers

pkg/apiserver/apiserver.go
```diff
@@ -80,10 +95,17 @@ func (c completedConfig) New(delegationTarget genericapiserver.DelegationTarget)
...
+			Config:                *c.GenericConfig.Config,
+			SharedInformerFactory: c.GenericConfig.SharedInformerFactory,
+			ClientConfig:          c.ClientConfig,
+		},
 		ExtraConfig: oauthapiserver.ExtraConfig{
 			// no one is allowed to set this today
 			ServiceAccountMethod: string(openshiftcontrolplanev1.GrantHandlerPrompt),
+
+			AccessTokenInactivityTimeout: c.ExtraConfig.AccessTokenInactivityTimeout,
+			APIAudiences:                 c.ExtraConfig.APIAudiences,
```
deads2k, line 108: ImplicitAudiences, right?

pkg/oauth/apiserver/apiserver.go
stlaz: :facepalm:

pkg/oauth/apiserver/apiserver.go
stlaz: :facepalm:

pkg/apiserver/apiserver.go
stlaz: upstream uses `APIAudiences`
https://github.com/openshift/kubernetes/blob/4b40bb4792d0b183fd02636660c55faac08a8735/pkg/kubeapiserver/authenticator/config.go#L183

deads2k:
/lgtm
/hold

holding for @sttts

pkg/oauth/apiserver/registry/tokenreviews/etcd.go
sttts: do we need the internal types?

sttts: The tokenreview `REST` object is self-contained, with exception of the internal types. So if we settled on v1, we could remove the dependency. Am happy with a follow-up to try that.

PR #14, opened 2020-07-27
configures the storage layer to be backward compatible with OpenShift API

p0lyn0mial:
this PR adds default resources prefixes for `oauth ` and `user` groups to be backward compatible with OpenShift API.

It also sets the default storage media type to `application/vnd.kubernetes.protobuf` to be backward compatible with OpenShift API


pkg/oauth/apiserver/apiserver.go
```diff
@@ -27,8 +25,8 @@ import (
 )
 
 var (
-	scheme = runtime.NewScheme()
-	codecs = serializer.NewCodecFactory(scheme)
+	scheme = legacyscheme.Scheme
```
p0lyn0mial, line 28: is it okay to refer to `legacyscheme.Scheme` ? Or do we prefer to plug a new one instead ?

pkg/cmd/oauth-apiserver/cmd.go
```diff
@@ -111,5 +116,46 @@ func (o *OAuthAPIServerOptions) NewOAuthAPIServer() (*apiserver.OAuthAPIServer,
 		return nil, err
 	}
 
+	etcdOptions := o.RecommendedOptions.Etcd
+	etcdOptions.DefaultStorageMediaType = "application/vnd.kubernetes.protobuf"
```
p0lyn0mial, line 120: this could be configured via the `storage-media-type` flag

pkg/cmd/oauth-apiserver/cmd.go
```diff
@@ -111,5 +116,46 @@ func (o *OAuthAPIServerOptions) NewOAuthAPIServer() (*apiserver.OAuthAPIServer,
...
+	if err := o.RecommendedOptions.Etcd.ApplyWithStorageFactoryTo(storageFactory, &serverConfig.GenericConfig.Config); err != nil {
+		return nil, err
+	}
+
 	return serverConfig.Complete().New(genericapiserver.NewEmptyDelegate())
 }
+
+// specialDefaultResourcePrefixes are a custom storage prefixes (we must be backward compatible with OpenShift API)
+var specialDefaultResourcePrefixes = map[schema.GroupResource]string{
+	{Resource: "oauthaccesstokens"}:                                 "oauth/accesstokens",
```
sttts, line 149: the non-grouped variants are probably not necessary anymore.

pkg/cmd/oauth-apiserver/cmd.go
```diff
@@ -111,5 +116,46 @@ func (o *OAuthAPIServerOptions) NewOAuthAPIServer() (*apiserver.OAuthAPIServer,
...
+	{Resource: "oauthauthorizetokens"}:                              "oauth/authorizetokens",
+	{Resource: "oauthauthorizetokens", Group: "oauth.openshift.io"}: "oauth/authorizetokens",
+
+	{Resource: "oauthclients"}:                                           "oauth/clients",
+	{Resource: "oauthclients", Group: "oauth.openshift.io"}:              "oauth/clients",
+	{Resource: "oauthclientauthorizations"}:                              "oauth/clientauthorizations",
+	{Resource: "oauthclientauthorizations", Group: "oauth.openshift.io"}: "oauth/clientauthorizations",
+
+	{Resource: "identities"}:                             "useridentities",
+	{Resource: "identities", Group: "user.openshift.io"}: "useridentities",
```
sttts, line 160: do we access these?

pkg/oauth/apiserver/apiserver.go
sttts: custom

pkg/cmd/oauth-apiserver/cmd.go
p0lyn0mial: the new server will manage `user.openshift.io`

pkg/cmd/oauth-apiserver/cmd.go
p0lyn0mial: okay, will remove.

p0lyn0mial:
/retitle configures the storage layer to be backward compatible with OpenShift API

pkg/cmd/oauth-apiserver/cmd.go
sttts: oh, right.

PR #23, opened 2020-10-20
bumps to kube 1.19.2 and picks up our carries to apiserver package

p0lyn0mial:


go.mod
```diff
@@ -33,7 +33,7 @@ replace (
 	k8s.io/api => k8s.io/api v0.19.2
 	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.19.2
 	k8s.io/apimachinery => k8s.io/apimachinery v0.19.2
-	k8s.io/apiserver => github.com/openshift/kubernetes-apiserver v0.0.0-20200727120649-a5e0645f4b46
+	k8s.io/apiserver => github.com/openshift/kubernetes-apiserver v0.0.0-20201019143247-401ead4ea686 // points to openshift-apiserver-4.7-kubernetes-1.19.2
```
p0lyn0mial, line 36: will be reused with openshift-apiserver

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/oauth-apiserver
 
-go 1.13
```
sttts, line 3: also in Dockerfile?

p0lyn0mial: yes, [the Dockerfile](https://github.com/openshift/oauth-apiserver/blob/master/images/Dockerfile.rhel7#L1) already points to `registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6`

sttts: We really need something in build machinery to check these, also in .ci-operator.yaml soon.

PR #24, opened 2020-10-21
allow users list their own access tokens

stlaz:
Adds an endpoint that mirrors the oauthaccesstokens that only belong to the requesting user. The new endpoint allows these operations:
- get
- list
- watch
- delete

cc @deads2k @sttts 

enhancement ref: https://github.com/openshift/enhancements/pull/507


pkg/oauth/apis/oauth/types.go
```diff
@@ -224,3 +224,19 @@ type RedirectReference struct {
 	Kind  string
 	Name  string
 }
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type PersonalAccessToken struct {
```
sttts, line 230: my vote goes to UserOauthAccessToken

pkg/oauth/apiserver/apiserver.go
```diff
@@ -139,11 +140,16 @@ func (c *completedConfig) newV1RESTStorage() (map[string]rest.Storage, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("error building REST storage: %v", err)
+	}
 
 	v1Storage := map[string]rest.Storage{}
 	v1Storage["oAuthAuthorizeTokens"] = authorizeTokenStorage
 	v1Storage["oAuthAccessTokens"] = accessTokenStorage
 	v1Storage["oAuthClients"] = clientStorage
 	v1Storage["oAuthClientAuthorizations"] = clientAuthorizationStorage
+	v1Storage["personalTokens"] = personalTokensDelegate
```
sttts, line 153: needs some naming unification following the enhancement.

pkg/oauth/apiserver/registry/personaltokens/delegate/delegate.go
```diff
@@ -0,0 +1,186 @@
...
+func (r *REST) New() runtime.Object {
+	return &oauthv1.PersonalAccessToken{}
+}
+
+func (r *REST) NewList() runtime.Object {
+	return &oauthv1.PersonalAccessTokenList{}
+}
+
+func (r *REST) List(ctx context.Context, options *metainternal.ListOptions) (runtime.Object, error) {
+	ctxUserName, err := getUserFromContext(ctx)
```
sttts, line 57: can it be empty?

pkg/oauth/apiserver/registry/personaltokens/delegate/delegate.go
```diff
@@ -0,0 +1,186 @@
...
+	if opts == nil {
+		newOpts = &metainternal.ListOptions{}
+	} else {
+		newOpts = opts.DeepCopy()
+	}
+
+	// the current field selector cannot make any assumptions over the "userName" field
+	for _, req := range newOpts.FieldSelector.Requirements() {
+		if req.Field == "userName" {
+			return nil, errors.NewUnauthorized("the userName field selector is not allowed")
```
sttts, line 177: can we alternatively just add our userName constraint? If the user adds another one for a different user, the set will just be empty. That's a cleaner and maybe safer solution.

sttts: You actually do that below. I.e. we could just drop this whole check?

pkg/oauth/apiserver/registry/personaltokens/delegate/delegate.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	tokenListGenericObj, err := r.accessTokenStorage.List(ctx, sanitizedListOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	tokenList := tokenListGenericObj.(*oauthapi.OAuthAccessTokenList)
```
sttts, line 72:
```
tokenList, ok := ...
if !ok {
```

The current code panics.

pkg/oauth/apiserver/registry/personaltokens/delegate/delegate.go
```diff
@@ -0,0 +1,186 @@
...
+	return r.tableConvertor.ConvertToTable(ctx, object, tableOptions)
+}
+
+func (r *REST) NamespaceScoped() bool {
+	return false
+}
+
+func (r *REST) Delete(ctx context.Context, name string, validateFunc rest.ValidateObjectFunc, options *metav1.DeleteOptions) (runtime.Object, bool, error) {
+	if _, err := r.Get(ctx, name, &metav1.GetOptions{}); err != nil {
+		return &metav1.Status{Status: metav1.StatusFailure}, false, err
```
sttts, line 138: here is a small risk for a race: you have to pass the uid of the returned object in the options below, and probably also do some translation of GetOptions to DeleteOptions above.

pkg/oauth/apiserver/registry/personaltokens/delegate/delegate.go
sttts: nvmd. It's your func below.

pkg/oauth/apiserver/registry/personaltokens/delegate/delegate.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	return r.accessTokenStorage.Delete(ctx, name, validateFunc, options)
+}
+
+func getUserFromContext(ctx context.Context) (string, error) {
+	userInfo, exists := apirequest.UserFrom(ctx)
+	if !exists {
+		return "", kerrors.NewForbidden(oauth.Resource("personaltokens"), "", fmt.Errorf("unable to get personal tokens without a user in context"))
```
sttts, line 147: Or just 404.

pkg/oauth/apiserver/registry/personaltokens/delegate/delegate.go
```diff
@@ -0,0 +1,186 @@
...
+		return "", kerrors.NewForbidden(oauth.Resource("personaltokens"), "", fmt.Errorf("user has a zero-length username"))
+	}
+
+	return userName, nil
+}
+
+func oauthAccessTokenListToUserOAuthAccessTokenList(l *oauthapi.OAuthAccessTokenList) *oauthapi.PersonalAccessTokenList {
+	ret := oauthapi.PersonalAccessTokenList{}
+	for _, t := range l.Items {
+		ret.Items = append(ret.Items, oauthapi.PersonalAccessToken{OAuthAccessToken: t})
```
sttts, line 161: do we use this pattern of embedding elsewhere? If not, I would be slightly concerned that things might break if TypeMeta and ObjectMeta are not at the toplevel of the type.

pkg/oauth/printers/internalversion/printer.go
```diff
@@ -159,6 +165,27 @@ func printOAuthAccessToken(oauthAccessToken *oauthapi.OAuthAccessToken, options
...
+		expires = created.Add(time.Duration(personalAccessToken.ExpiresIn) * time.Second).String()
+	}
+	row.Cells = append(row.Cells,
+		personalAccessToken.Name,
+		personalAccessToken.UserName,
+		personalAccessToken.ClientName,
+		created,
+		expires,
+		personalAccessToken.RedirectURI,
+		strings.Join(personalAccessToken.Scopes, ","),
```
sttts, line 184: @Anandnatraj these are the columns with oc.

pkg/oauth/apiserver/registry/personaltokens/delegate/delegate.go
stlaz: `getUserFromContext` checks for an empty username and errors out in that case

pkg/oauth/apiserver/registry/personaltokens/delegate/delegate.go
stlaz: I assume it's novel, I wasn't looking for it elsewhere

sttts: we will see. If you have it working, maybe everything is fine.

stlaz:
/retest
I wonder why it's a baremetal operator that failed on e2e-aws

stlaz:
Addressed all comments. The PR now depends on https://github.com/openshift/api/pull/774 (aside from the enhancement merging)

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,216 @@
...
+)
+
+// rest implements a RESTStorage for access tokens against etcd
+type REST struct {
+	accessTokenStorage *accesstokenregistry.REST
+
+	tableConvertor rest.TableConvertor
+}
+
+var _ rest.Lister = &REST{}
```
deads2k, line 36: isn't there a standard REST you can match against?

deads2k: > isn't there a standard REST you can match against?

ah, no create or update.  Add comment.

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,216 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	token := tokenGenericObj.(*oauthapi.OAuthAccessToken)
+	if token == nil {
+		return nil, errors.NewInternalError(fmt.Errorf("failed to convert generic accesstoken GET result to its typed version"))
+	}
+
+	if token.UserName != ctxUserName || !isTokenSha256Token(token.Name) { // don't reveal other people tokens' hashes or non-hashed tokens
```
deads2k, line 98: make this a filter func so you can re-use it everywhere.

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,216 @@
...
+	if options.Preconditions != nil && options.Preconditions.ResourceVersion != nil {
+		deletedRV = *options.Preconditions.ResourceVersion
+	}
+
+	deletedObj, err := r.Get(ctx, name, &metav1.GetOptions{TypeMeta: options.TypeMeta, ResourceVersion: deletedRV})
+	if err != nil {
+		return &metav1.Status{Status: metav1.StatusFailure}, false, err
+	}
+
+	deletedTyped := deletedObj.(*oauthapi.UserOAuthAccessToken)
```
deads2k, line 145: check user

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,216 @@
...
+		return "", kerrors.NewNotFound(oauth.Resource("useroauthaccesstokens"), "")
+	}
+
+	return userName, nil
+}
+
+func oauthAccessTokenListToUserOAuthAccessTokenList(l *oauthapi.OAuthAccessTokenList) *oauthapi.UserOAuthAccessTokenList {
+	ret := oauthapi.UserOAuthAccessTokenList{}
+	for _, t := range l.Items {
+		if !isTokenSha256Token(t.Name) {
```
deads2k, line 185: confirm user just in case

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/proxywatcher.go
```diff
@@ -0,0 +1,106 @@
...
+			case watch.Error:
+				w.outgoing <- event
+			default:
+				tokenOrig, ok := event.Object.(*oauthapi.OAuthAccessToken)
+				if !ok {
+					w.outgoing <- createErrorEvent(errors.NewInternalError(fmt.Errorf("failed to convert incoming object to an OAuthAccessToken type")))
+					continue
+				}
+
+				if !isTokenSha256Token(tokenOrig.Name) {
```
deads2k, line 88: filter by name too.

Also, I think there's a first class filter concept.

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
stlaz: It's checked in the `Get()` method but I agree that explicit is better than implicit and who knows what can happen to the `Get()` in the future. I added that check.

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/proxywatcher.go
stlaz: Added the filtering.

Not sure whether there's a better, first class filtering. The channel I'm "borrowing" belongs to another watcher that seems to only allow conventional ListOptions

stlaz:
/hold cancel
this is not a prototype anymore

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,228 @@
...
+func userFieldSelectorToListOptions(opts *metainternal.ListOptions, userName string) (*metainternal.ListOptions, error) {
+	var newOpts *metainternal.ListOptions
+	if opts == nil {
+		newOpts = &metainternal.ListOptions{}
+	} else {
+		newOpts = opts.DeepCopy()
+	}
+
+	newOpts.FieldSelector = fields.AndSelectors(
+		fields.Set{"userName": userName}.AsSelector(),
```
sttts, line 217: we need an e2e test for this

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,228 @@
...
+			continue
+		}
+		ret.Items = append(ret.Items, oauthapi.UserOAuthAccessToken(t))
+	}
+
+	// when watching a resource, the client usually attempts to list it first and
+	// it then continues to watch from the last known RV
+	// This prevents the initial objects to be returned twice to the client
+	// (objects from the list + from the next watch with an uninitialized from-RV)
+	ret.ResourceVersion = l.ResourceVersion
```
sttts, line 204: why not everything else from the meta fields?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	var deletedRV string
+	if options.Preconditions != nil && options.Preconditions.ResourceVersion != nil {
+		deletedRV = *options.Preconditions.ResourceVersion
+	}
+
+	deletedObj, err := r.Get(ctx, name, &metav1.GetOptions{TypeMeta: options.TypeMeta, ResourceVersion: deletedRV})
+	if err != nil {
+		return &metav1.Status{Status: metav1.StatusFailure}, false, err
```
sttts, line 148: why not nil as object?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,228 @@
...
+	// don't mutate the original options
+	newOpts := options.DeepCopy()
+
+	switch {
+	case newOpts.Preconditions == nil:
+		newOpts.Preconditions = &metav1.Preconditions{UID: &deletedTyped.UID}
+
+	case newOpts.Preconditions != nil && newOpts.Preconditions.UID != nil && *newOpts.Preconditions.UID != deletedTyped.UID:
+		// the UID seeked to be deleted no longer exists/never existed
+		return &metav1.Status{Status: metav1.StatusFailure}, false, errors.NewNotFound(oauth.Resource("useroauthaccesstokens"), name)
```
sttts, line 165: wouldn't we get this for free from the delete call below?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,228 @@
...
+		newOpts.Preconditions.UID = &deletedTyped.UID
+	}
+
+	return r.accessTokenStorage.Delete(ctx, name, validateFunc, newOpts)
+}
+
+func getUserFromContext(ctx context.Context) (string, error) {
+	userInfo, exists := apirequest.UserFrom(ctx)
+	if !exists {
+		klog.V(4).Infof("no user found in the context")
```
sttts, line 178: do we want this? When can it happen?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/proxywatcher.go
```diff
@@ -0,0 +1,112 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+
+	oauthapi "github.com/openshift/oauth-apiserver/pkg/oauth/apis/oauth"
+)
+
+// OAuthAccessTokenWatcher implements a Watcher interface to watch OAuthAccessTokens
+// and filter them to match only those for the current user
+type OAuthAccessTokenWatcher struct {
```
sttts, line 17: is this taken from anywhere?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/proxywatcher.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+}
+
+// Stop implements Interface
+func (w *OAuthAccessTokenWatcher) Stop() {
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+	if !w.stopped {
+		w.stopped = true
+		// forward the Stop() request
```
sttts, line 52: no additional value of the comment

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
stlaz: I added some e2e tests here: https://github.com/openshift/origin/pull/25697

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
stlaz: I thought I would limit the amount of information to only that we really need. I however understand that it is hard to objectively tell which information is useful and necessary and which is not, I think I'll add all of the metadata, although perhaps conditionally when at least _some_ items appear in the filtered slice?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
stlaz: I was not sure how exactly this should be handled and the docs of the `GracefulDeleter` made me think this was the right way, but reading the code of the generic `Store`, it seems that `nil` would be the right option.

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
stlaz: I am not sure whether it's at all possible, it felt like a safe precaution.

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/proxywatcher.go
stlaz: I was looking for inspiration in the watchers of k8s.io/apimachinery/pkg/watch but I ended up mixing some of the basic bits of the openshift-apiserver's project watcher with custom resource proxying logic

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
stlaz: actually, never mind, I think we can always set it

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
stlaz: I think so... I wanted to prevent user-supplied options being passed to the actual privileged call, but given we make sure the preconditions/name combination does not exist, we should be able to get to the same result.

stlaz:
The latest revision adds fixes that:
- prevent panic on fieldselector being nil
- added `clientName` field selector for the new API type
- fixed an issue preventing Get from working correctly

Todo:
- move e2e tests here

stlaz:
/hold
now depends on https://github.com/openshift/api/pull/788 and https://github.com/openshift/client-go/pull/167

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
stlaz: moved e2e tests to this repo

stlaz:
/hold cancel
removed fake bumps as dependencies merged
https://github.com/openshift/release/pull/13807 adds the ci configuration for the component e2e tests

stlaz:
rbac for the new endpoints added by https://github.com/openshift/cluster-authentication-operator/pull/381

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,218 @@
...
+	oauthv1 "github.com/openshift/api/oauth/v1"
+
+	oauthapi "github.com/openshift/oauth-apiserver/pkg/oauth/apis/oauth"
+	accesstokenregistry "github.com/openshift/oauth-apiserver/pkg/oauth/apiserver/registry/oauthaccesstoken/etcd"
+	oauthprinters "github.com/openshift/oauth-apiserver/pkg/oauth/printers/internalversion"
+	"github.com/openshift/oauth-apiserver/pkg/printers"
+	"github.com/openshift/oauth-apiserver/pkg/printerstorage"
+)
+
+// rest implements a RESTStorage for access tokens against etcd
```
sttts, line 29: nit: for access tokens that the current user owns (selected by label)

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
sttts: Would remove the log output here.

sttts: or even turn this into `string, bool, error` and return the NotFound in the call sites.

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,218 @@
...
+func getUserFromContext(ctx context.Context) (string, error) {
+	userInfo, exists := apirequest.UserFrom(ctx)
+	if !exists {
+		klog.V(4).Infof("no user found in the context")
+		return "", kerrors.NewNotFound(oauth.Resource("useroauthaccesstokens"), "")
+	}
+
+	userName := userInfo.GetName()
+	if len(userName) == 0 {
+		klog.V(4).Info("user in context has zero-length username")
```
sttts, line 176: same here. No username means nothing to see. `return "", false, nil` and NotFound where this func is called.

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,218 @@
...
+			continue
+		}
+		ret.Items = append(ret.Items, oauthapi.UserOAuthAccessToken(t))
+	}
+
+	ret.ListMeta = l.ListMeta
+	return &ret
+}
+
+func userFieldSelectorToListOptions(opts *metainternal.ListOptions, userName string) (*metainternal.ListOptions, error) {
```
sttts, line 196: ListOptionsWithUserNameFilter

test/e2e/useroauthaccesstokens_test.go
```diff
@@ -0,0 +1,560 @@
...
+			userToken:       frantaTokenString,
+			userName:        "franta",
+			fieldSelector:   fields.OneTermEqualSelector("clientName", "openshift-browser-client"),
+			expectedResults: 2,
+		},
+		{
+			name:            "set label selector to get own tokens and of others",
+			userToken:       mirkaTokenString,
+			userName:        "mirka",
+			labelSelector:   parseLabelSelectorOrDie("randomLabel notin (mirka,franta)"),
```
sttts, line 255: would add the simple case too: userName == somebody else.

test/e2e/useroauthaccesstokens_test.go
```diff
@@ -0,0 +1,560 @@
...
+			name:          "delete someone else's token",
+			userToken:     mirkaTokenString,
+			getTokenName:  "sha256~" + testNS + "franta0",
+			expectedError: true,
+		},
+		{
+			name:         "delete an own token",
+			userToken:    frantaTokenString,
+			userName:     "mirka",
+			getTokenName: "sha256~" + testNS + "franta0",
```
sttts, line 368: also test with a wrong UID.

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
stlaz: what's the use of error, then?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	sanitizedListOpts, err := userFieldSelectorToListOptions(options, ctxUserName)
+	if err != nil {
+		return nil, err
+	}
+
+	tokenListGenericObj, err := r.accessTokenStorage.List(ctx, sanitizedListOpts)
+	if err != nil {
+		return nil, err
```
p0lyn0mial, line 72: should we mask the err ?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	tokenList, ok := tokenListGenericObj.(*oauthapi.OAuthAccessTokenList)
+	if !ok {
+		return nil, errors.NewInternalError(fmt.Errorf("failed to convert generic accesstoken LIST result to its typed version"))
```
p0lyn0mial, line 77: could we be more specific and log the underlying type for `tokenListGenericObj` and `OAuthAccessTokenList` ?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,218 @@
...
+
+func userFieldSelectorToListOptions(opts *metainternal.ListOptions, userName string) (*metainternal.ListOptions, error) {
+	var newOpts *metainternal.ListOptions
+	if opts == nil {
+		newOpts = &metainternal.ListOptions{}
+	} else {
+		newOpts = opts.DeepCopy()
+	}
+
+	userSelector := fields.Set{"userName": userName}.AsSelector()
```
p0lyn0mial, line 204: what if `opts` already specifies the field selector for the `userName`?

I think we should be fine since we have an additional check `isValidUserToken` but maybe we could forbid specifying that particular selector?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
p0lyn0mial, line 196: this method doesn't return `error` - drop it.

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/proxywatcher.go
```diff
@@ -0,0 +1,111 @@
...
+					continue
+				}
+
+				if !isValidUserToken(tokenOrig, w.username) {
+					continue
+				}
+				event.Object = (*oauthapi.UserOAuthAccessToken)(tokenOrig)
+				w.outgoing <- event
+			}
+		case <-ctx.Done(): // user cancel
```
p0lyn0mial, line 99: in that case, we won't set `w.stopped = true` and won't close the  `close(w.stopCh)`

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/proxywatcher.go
```diff
@@ -0,0 +1,111 @@
...
+	return w.stopped
+}
+
+// ResultChan returns channel for filtered results
+func (w *OAuthAccessTokenWatcher) ResultChan() <-chan watch.Event {
+	return w.outgoing
+}
+
+// StopChan returns stop channel
+func (w *OAuthAccessTokenWatcher) StopChan() <-chan struct{} {
```
p0lyn0mial, line 70: is this method used ?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/proxywatcher.go
```diff
@@ -0,0 +1,111 @@
...
+	defer w.mutex.Unlock()
+	if !w.stopped {
+		w.stopped = true
+		w.wrappedWatcher.Stop()
+		close(w.stopCh)
+	}
+}
+
+// Stopping returns true if Stop() has been called
+func (w *OAuthAccessTokenWatcher) Stopping() bool {
```
p0lyn0mial, line 58: is this method used?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	tokenList, ok := tokenListGenericObj.(*oauthapi.OAuthAccessTokenList)
+	if !ok {
+		return nil, errors.NewInternalError(fmt.Errorf("failed to convert generic accesstoken LIST result to its typed version"))
+	}
+
+	return oauthAccessTokenListToUserOAuthAccessTokenList(tokenList, ctxUserName), nil
```
p0lyn0mial, line 80: at this point `tokenList != nil`, right ?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
sttts: right, just get rid of it. I prefer to put logic for errors returned to the user at the top-level, not deep into helpers.

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
sttts: we call `AndSelectors` which will make sure our user is always checked for. If the user checks for another one, the selector will always be false. That's what we want. Compare e2e tests. There is one for that case.

test/e2e/useroauthaccesstokens_test.go
stlaz: there's a fieldSelector like this, but I'll add a label to one of the basic tokens above and will attempt to list it

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
stlaz: I think we *should* be fine with the error as is

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
stlaz: This error gets propagated to the user, does it not? I think it might be confusing when they see a mention of a different type from what they requested

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/proxywatcher.go
stlaz: good catch

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/proxywatcher.go
stlaz: I copied the interface of the `apimachinery/pkg/watch.ProxyWatcher` struct, but it looks like this and `Stopping` may indeed be safe to remove

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
stlaz: yes, that should hold

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -57,15 +55,12 @@ func (r *REST) NewList() runtime.Object {
 }
 
 func (r *REST) List(ctx context.Context, options *metainternal.ListOptions) (runtime.Object, error) {
-	ctxUserName, err := getUserFromContext(ctx)
-	if err != nil {
-		return nil, err
+	ctxUserName, ok := getUserFromContext(ctx)
+	if !ok {
+		return nil, errors.NewNotFound(oauth.Resource("useroauthaccesstokens"), "")
```
sttts, line 60: Shouldn't this not rather be just an empty result?

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -81,18 +76,18 @@ func (r *REST) List(ctx context.Context, options *metainternal.ListOptions) (run
 }
 
 func (r *REST) Get(ctx context.Context, name string, options *metav1.GetOptions) (runtime.Object, error) {
-	ctxUserName, err := getUserFromContext(ctx)
-	if err != nil {
-		return nil, err
+	ctxUserName, ok := getUserFromContext(ctx)
+	if !ok {
+		return nil, errors.NewNotFound(oauth.Resource("useroauthaccesstokens"), "")
```
sttts, line 81: as for List, but with second argument to be name

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
```diff
@@ -133,9 +125,9 @@ func (r *REST) NamespaceScoped() bool {
 }
 
 func (r *REST) Delete(ctx context.Context, name string, validateFunc rest.ValidateObjectFunc, options *metav1.DeleteOptions) (runtime.Object, bool, error) {
-	ctxUserName, err := getUserFromContext(ctx)
-	if err != nil {
-		return nil, false, errors.NewNotFound(oauth.Resource("useroauthaccesstokens"), name)
+	ctxUserName, ok := getUserFromContext(ctx)
+	if !ok {
+		return nil, false, errors.NewNotFound(oauth.Resource("useroauthaccesstokens"), "")
```
sttts, line 130: keep the name

pkg/oauth/apiserver/registry/useroauthaccesstokens/delegate/delegate.go
p0lyn0mial: When the conversion can fail? Can it only happen when the wiring is wrong? 

In that case, I would prefer to see a very clear and specific error so that I can debug it.

stlaz:
The component tests are currently failing because https://github.com/openshift/cluster-authentication-operator/pull/381 hasn't merged yet

stlaz:
/hold
wait for rbac deps to merge

stlaz:
we still need https://github.com/openshift/cluster-authentication-operator/pull/381 for the tests to pass :/

stlaz:
/hold cancel
the operator PR merged. We'll probably need a few retests before the RBAC is actually there

PR #40, opened 2021-02-09
completeConfig: remove storage, storageErr state from config

sttts:
Cleanup in preparation of token review support.

### DIFF
--- a/pkg/oauth/apiserver/apiserver.go
+++ b/pkg/oauth/apiserver/apiserver.go
@@ -2,7 +2,6 @@ package apiserver
 
 import (
 	"fmt"
-	"sync"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/oauth/apiserver/apiserver.go
```diff
@@ -86,14 +82,6 @@ func (c completedConfig) New(delegationTarget genericapiserver.DelegationTarget)
 	return s, nil
 }
 
-func (c *completedConfig) V1RESTStorage() (map[string]rest.Storage, error) {
```
stlaz, line 89: Since we're getting rid of it (yay \o/), please also remove the same code in the user apiserver

stlaz (changes requested):

pkg/oauth/apiserver/apiserver.go
sttts: done

PR #41, opened 2021-02-11
add a custom authorizer to ensure kube-apiserver can always access tokenreviews webhook

deads2k:
Adds a hardcoded authorizer to avoid unnecessary SAR calls.  This could be made generic.  We could also add one for handling /metrics

### DIFF
--- a/pkg/authorization/hardcodedauthorizer/tokenreview.go
+++ b/pkg/authorization/hardcodedauthorizer/tokenreview.go
@@ -1,0 +1,41 @@
+package hardcodedauthorizer
+
+import (
+	"context"
+
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+)
+
+type tokenReviewAuthorizer struct{}
+
+// GetUser() user.Info - checked
+// GetVerb() string - checked
+// IsReadOnly() bool - na
+// GetNamespace() string - checked
+// GetResource() string - checked
+// GetSubresource() string - checked
+// GetName() string - na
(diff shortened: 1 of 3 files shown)


deads2k:
/hold

This demonstrates the how. It would need unit tests at least.

pkg/authorization/tokenreviewauthorizer/tokenreviewauthorizer.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"context"
+
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+)
+
+type tokenReviewAuthorizer struct{}
+
+func (tokenReviewAuthorizer) Authorize(ctx context.Context, a authorizer.Attributes) (authorized authorizer.Decision, reason string, err error) {
+	if a.GetUser().GetName() != "system:serviceaccount:openshift-authentication:tokenreviewer"{
```
p0lyn0mial, line 12: alright, so this is a service account token used by the KAS, right?

deads2k: > alright, so this is a service account token used by the KAS, right?

In @stlaz's PR, yes.

p0lyn0mial: Got it. This is just an example. Thanks for sharing it with me.

lgtm

PR #44, opened 2021-02-24
forbid the old token format

stlaz:
let's watch the world burn

### DIFF
--- a/pkg/oauth/apis/oauth/validation/validation.go
+++ b/pkg/oauth/apis/oauth/validation/validation.go
@@ -33,6 +33,7 @@ const (
 const (
 	codeChallengeMethodPlain  = "plain"
 	codeChallengeMethodSHA256 = "S256"
+	sha256Prefix              = "sha256~"
 )
 
 var CodeChallengeMethodsSupported = []string{codeChallengeMethodPlain, codeChallengeMethodSHA256}
@@ -44,6 +45,10 @@ func ValidateTokenName(name string, prefix bool) []string {
 
 	if len(name) < MinTokenLength {
 		return []string{fmt.Sprintf("must be at least %d characters long", MinTokenLength)}
+	}
+
+	if !strings.HasPrefix(name, sha256Prefix) {
+		return []string{fmt.Sprintf("only the new format of tokens (prefixed with sha256~) is allowed")}
 	}
 	return nil
 }
--- a/pkg/tokenvalidation/bootstrapauthenticator.go
+++ b/pkg/tokenvalidation/bootstrapauthenticator.go
@@ -45,7 +45,7 @@ func (a *bootstrapAuthenticator) AuthenticateToken(ctx context.Context, name str
 		name = sha256Prefix + base64.RawURLEncoding.EncodeToString(h[0:])
 	}
 
-	token, err := a.tokens.Get(context.TODO(), name, metav1.GetOptions{})
+	token, err := a.tokens.Get(ctx, name, metav1.GetOptions{})
 	if err != nil {
 		return nil, false, errLookup // mask the error so we do not leak token data in logs
 	}
--- a/pkg/tokenvalidation/bootstrapauthenticator.go
+++ b/pkg/tokenvalidation/bootstrapauthenticator.go
@@ -37,13 +37,13 @@ func NewBootstrapAuthenticator(tokens oauthclient.OAuthAccessTokenInterface, get
 }
 
 func (a *bootstrapAuthenticator) AuthenticateToken(ctx context.Context, name string) (*kauthenticator.Response, bool, error) {
-	// hash token for new-style sha256~ prefixed token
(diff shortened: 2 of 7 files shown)


sttts:
@stlaz how can we make sure we don't break customers? We need a release-note for sure. But we probably also want some kind of alert in 4.7 ? cc @bergerhoffer

bergerhoffer:
> @stlaz how can we make sure we don't break customers? We need a release-note for sure. But we probably also want some kind of alert in 4.7 ? cc @bergerhoffer

@sttts If we're disallowing the old format tokens starting with 4.8 then yeah I think we'd need to mention something in the docs earlier than 4.8 as an FYI for customers. I don't know whether a z-stream note wold be sufficient, but we can check. An alert also sounds like a good idea so that people aren't surprised.

EmilyM1:
This ok to restest?

stlaz:
> This ok to restest?

until https://github.com/openshift/origin/pull/25965 merges, the e2e tests are likely to fail

stlaz:
/retest
the origin PR merged, let's see what else we've got

stlaz:
/test e2e-aws-serial
Looks like there are no token-related failures in the serial test suite, rerunning to make sure

stlaz:
https://github.com/openshift/origin/pull/26009 should fix all the remaining global-e2e tests issues

stlaz:
/retest
the origin PR merged

stlaz:
/retest
unrelated

pkg/oauth/apis/oauth/validation/validation.go
```diff
@@ -45,6 +46,10 @@ func ValidateTokenName(name string, prefix bool) []string {
 	if len(name) < MinTokenLength {
 		return []string{fmt.Sprintf("must be at least %d characters long", MinTokenLength)}
 	}
+
+	if !strings.HasPrefix(name, sha256Prefix) {
+		return []string{fmt.Sprintf("only the new format of tokens (prefixed with sha256~) is allowed")}
+	}
```
sttts, line 52: will this update of existing tokens?

stlaz: Current validation does not allow updating anything else but the `inactivityTimeoutSeconds`, so existing tokens will not be updated. The `inactivityTimeoutSeconds` of existing tokens will no longer be possible to be updated, so those will just time out.

Other, long-lived pre-existing tokens may still survive.

sttts: ack

pkg/tokenvalidation/bootstrapauthenticator.go
```diff
@@ -37,14 +38,14 @@ func NewBootstrapAuthenticator(tokens oauthclient.OAuthAccessTokenInterface, get
...
 
 func (a *bootstrapAuthenticator) AuthenticateToken(ctx context.Context, name string) (*kauthenticator.Response, bool, error) {
-	// hash token for new-style sha256~ prefixed token
-	// TODO: reject non-sha256 prefix tokens in 4.7+
-	if strings.HasPrefix(name, sha256Prefix) {
-		withoutPrefix := strings.TrimPrefix(name, sha256Prefix)
-		h := sha256.Sum256([]byte(withoutPrefix))
-		name = sha256Prefix + base64.RawURLEncoding.EncodeToString(h[0:])
+	if !strings.HasPrefix(name, sha256Prefix) {
+		errOldFormat = errors.New("old token format")
```
sttts, line 42: `return nil, false, errOldFormat` ?

pkg/oauth/apis/oauth/validation/validation.go
stlaz: btw: the newest changes make the old tokens disfunctional for authn

pkg/tokenvalidation/tokenauthenticator.go
```diff
@@ -17,7 +17,10 @@ import (
 	userclient "github.com/openshift/client-go/user/clientset/versioned/typed/user/v1"
 )
 
-var errLookup = errors.New("token lookup failed")
+var (
+	errLookup    = errors.New("token lookup failed")
+	errOldFormat = errors.New("old token format")
```
sttts, line 22: would rephrase this like "old, insecure token format" ?

pkg/oauth/apis/oauth/validation/validation.go
```diff
@@ -45,6 +46,10 @@ func ValidateTokenName(name string, prefix bool) []string {
 	if len(name) < MinTokenLength {
 		return []string{fmt.Sprintf("must be at least %d characters long", MinTokenLength)}
 	}
+
+	if !strings.HasPrefix(name, sha256Prefix) {
+		return []string{fmt.Sprintf("only the new format of tokens (prefixed with sha256~) is allowed")}
```
mfojtik, line 51: drop `fmt.Sprintf` and just use string

PR #47, opened 2021-03-30
loosen up username and identity name validation

stlaz:
Allows to include ':' in usernames if prepended by 'b64'.
This is to allow creating identities/users whose IdP-side
identifier contains either ':' or '/'. These are base64-encoded
and the result is appended after the 'b64:' tag in the object
name.

cc @sttts @deads2k


stlaz:
/hold
contains fake bump

stlaz:
/retest
irrelevant

pkg/user/apis/user/validation/validation.go
```diff
@@ -18,13 +18,20 @@ func ValidateIdentityName(name string, _ bool) []string {
 	}
 
 	parts := strings.Split(name, ":")
+	var providerIndex, usernameIndex = 0, 1
 	if len(parts) != 2 {
-		return []string{`must be in the format <providerName>:<providerUserName>`}
+		// allow identities in the form "<providerName>:b64:<base64-encoded username>"
+		if len(parts) == 3 && parts[1] == "b64" {
+			providerIndex++
```
sttts, line 25: why providerIndex++ ?

stlaz:
/hold cancel
no longer contains fake bump

pkg/user/apis/user/validation/validation.go
```diff
@@ -18,13 +18,19 @@ func ValidateIdentityName(name string, _ bool) []string {
 	}
 
 	parts := strings.Split(name, ":")
+	usernameIndex := 1
 	if len(parts) != 2 {
-		return []string{`must be in the format <providerName>:<providerUserName>`}
+		// allow identities in the form "<providerName>:b64:<base64-encoded username>"
+		if len(parts) == 3 && parts[1] == "b64" {
```
sttts, line 24: would be easier to read through
```
if len(parts) == 3 && parts[1] == "b64" {
 ...
} else if len(parts) != 2 { 
 ...
}
```

sttts:
/retest
/lgtm
/approve

Feel free to fix the nit as the long tests have to rerun anyway.

PR #48, opened 2021-04-06
prints flags at startup

p0lyn0mial:
since we pass flags directly from the operator to binary having them in logs at startup is helpful.


p0lyn0mial:
xref: https://github.com/openshift/cluster-authentication-operator/pull/429

PR #52, opened 2021-05-04
Bug 1958097: tokenreviews: improve errors on old-format access tokens

stlaz:
Neither the old format nor serviceaccount tokens have the sha256~
prefix, return a NotFound error in case the tokenvalidation endpoint
is reached with a non-prefixed token.

/assign @sttts 
/cc @s-urbaniak

### DIFF
--- a/pkg/tokenvalidation/bootstrapauthenticator.go
+++ b/pkg/tokenvalidation/bootstrapauthenticator.go
@@ -38,7 +38,13 @@ func NewBootstrapAuthenticator(tokens oauthclient.OAuthAccessTokenInterface, get
 
 func (a *bootstrapAuthenticator) AuthenticateToken(ctx context.Context, name string) (*kauthenticator.Response, bool, error) {
 	if !strings.HasPrefix(name, sha256Prefix) {
-		return nil, false, errOldFormat
+		// only complain about non-sha256 format if the token is really an existing
+		// OAuthAccessToken (and e.g. no service account which also has no sha256 prefix)
+		_, err := a.tokens.Get(ctx, name, metav1.GetOptions{})
+		if err == nil {
+			return nil, false, errOldFormat
+		}
+		return nil, false, errLookup
 	}
 
 	withoutPrefix := strings.TrimPrefix(name, sha256Prefix)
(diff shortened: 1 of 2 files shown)


stlaz:
/bugzilla refresh
.retest

pkg/tokenvalidation/bootstrapauthenticator.go
```diff
@@ -38,7 +38,12 @@ func NewBootstrapAuthenticator(tokens oauthclient.OAuthAccessTokenInterface, get
 
 func (a *bootstrapAuthenticator) AuthenticateToken(ctx context.Context, name string) (*kauthenticator.Response, bool, error) {
 	if !strings.HasPrefix(name, sha256Prefix) {
-		return nil, false, errOldFormat
+		_, err := a.tokens.Get(ctx, name, metav1.GetOptions{})
```
sttts, line 41: a comment please for thiis block.

pkg/tokenvalidation/bootstrapauthenticator.go
```diff
@@ -38,7 +38,12 @@ func NewBootstrapAuthenticator(tokens oauthclient.OAuthAccessTokenInterface, get
...
+		if err == nil {
+			return nil, false, errOldFormat
+		} else {
```
sttts, line 44: remove else parenthesis

stlaz: :+1: how very silly of me

pkg/tokenvalidation/bootstrapauthenticator.go
```diff
@@ -38,7 +38,12 @@ func NewBootstrapAuthenticator(tokens oauthclient.OAuthAccessTokenInterface, get
 
 func (a *bootstrapAuthenticator) AuthenticateToken(ctx context.Context, name string) (*kauthenticator.Response, bool, error) {
 	if !strings.HasPrefix(name, sha256Prefix) {
-		return nil, false, errOldFormat
+		// check that the token exists and report it's of an insecure format if it does
```
sttts, line 41: would prefer something like:
```
// only complain about non-sha256 format if the token is really an existing OAuthAccessToken (and e.g. no service account which also has no sha256 prefix)
```

PR #54, opened 2021-06-08
Bug 1955435: Do not validate kube:admin user

slaskawi:
https://bugzilla.redhat.com/show_bug.cgi?id=1955435

This Pull Request prevents `kube:admin` user from failing with validation. 

Steps to verify this patch on a running server:
- Option 1 - observe the audit logs in the oAuth API Server logs and log in as `kubeadmin`. Note, there's no audit error emitted.
- Option 2 - use the following command and observe the output (note, there's no error):

```
$ oc get --raw /apis/user.openshift.io/v1/users/kube:admin
Error from server (NotFound): users.user.openshift.io "kube:admin" not found
```

### DIFF
--- a/pkg/oauth/apis/oauth/validation/validation_test.go
+++ b/pkg/oauth/apis/oauth/validation/validation_test.go
@@ -1,6 +1,8 @@
 package validation
 
 import (
+	"reflect"
+	"strings"
 	"testing"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -744,5 +746,33 @@ func TestValidateAuthorizeTokensUpdate(t *testing.T) {
 				t.Errorf("%s: expected errors to have field %s: %v", k, v.F, errs[i])
 			}
 		}
+	}
+}
+
+func TestValidateUserNameField(t *testing.T) {
+	fieldPath := field.NewPath("metadata", "name")
+	tests := []struct {
+		username string
+		want     field.ErrorList
+	}{
+		{
+			username: "john",
+			want:     field.ErrorList{},
+		},
+		{
+			username: "kube:admin",
+			want:     field.ErrorList{},
+		},
+		{
+			username: "system:admin",
+			want:     field.ErrorList{field.Invalid(fieldPath, "system:admin", strings.Join([]string{`usernames that contain ":" must begin with "b64:"`}, ", "))},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.username, func(t *testing.T) {
+			if got := ValidateUserNameField(tt.username, fieldPath); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("ValidateUserNameField() = %v, want %v", got, tt.want)
+			}
+		})
 	}
 }
--- a/pkg/user/apiserver/registry/user/etcd/etcd.go
+++ b/pkg/user/apiserver/registry/user/etcd/etcd.go
@@ -3,8 +3,6 @@ package etcd
 import (
 	"context"
 	"errors"
-	"strings"
-
 	kerrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -18,6 +16,7 @@ import (
 
 	usergroup "github.com/openshift/api/user"
 	"github.com/openshift/apiserver-library-go/pkg/apivalidation"
+	"github.com/openshift/oauth-apiserver/pkg/oauth/apis/oauth/validation"
 	"github.com/openshift/oauth-apiserver/pkg/printers"
 	"github.com/openshift/oauth-apiserver/pkg/printerstorage"
 	userapi "github.com/openshift/oauth-apiserver/pkg/user/apis/user"
@@ -101,9 +100,8 @@ func (r *REST) Get(ctx context.Context, name string, options *metav1.GetOptions)
 
 	// do not bother looking up users that cannot be persisted
 	// make sure we return a status error otherwise the API server will complain
-	if reasons := apivalidation.ValidateUserName(name, false); len(reasons) != 0 {
-		err := field.Invalid(field.NewPath("metadata", "name"), name, strings.Join(reasons, ", "))
-		return nil, kerrs.NewInvalid(usergroup.Kind("User"), name, field.ErrorList{err})
+	if errorList := validation.ValidateUserNameField(name, field.NewPath("metadata", "name")); len(errorList) != 0 {
+		return nil, kerrs.NewInvalid(usergroup.Kind("User"), name, errorList)
 	}
 
 	return r.Store.Get(ctx, name, options)


pkg/user/apiserver/registry/user/etcd/etcd.go
```diff
@@ -101,10 +100,16 @@ func (r *REST) Get(ctx context.Context, name string, options *metav1.GetOptions)
...
 
 	return r.Store.Get(ctx, name, options)
 }
+
+func validateUsername(name string) *kerrs.StatusError {
+	if errorList := validation.ValidateUserNameField(name, field.NewPath("metadata", "name")); len(errorList) != 0 {
+		return kerrs.NewInvalid(usergroup.Kind("User"), name, errorList)
+	}
+	return nil
+}
```
stlaz, line 115: Inline it back, please. I don't see a reason to keep this outside.

pkg/user/apiserver/registry/user/etcd/etcd_test.go
```diff
@@ -0,0 +1,30 @@
...
+	require.EqualValues(t, 422, err.Status().Code)
+}
+
+func TestValidUserValidation(t *testing.T) {
+	userToBeTested := "john"
+
+	err := validateUsername(userToBeTested)
+
+	require.Nil(t, err)
+}
```
stlaz, line 30: this is basically just testing the `ValidateUserNameField` function behavior, we don't need to do that here. Feel free to move this to the same module, use golang generated tests to do that (this should be only one `Test*` function)

stlaz (changes requested):

pkg/user/apiserver/registry/user/etcd/etcd_test.go
s-urbaniak: consider using table driven tests, i.e. dave cheney once write a nice blog post: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

pkg/user/apiserver/registry/user/etcd/etcd.go
slaskawi: I've extracted the validation part to make it unit-testable. Otherwise, I'd need to bootstrap the whole REST server.

slaskawi: Let me get this back. I'll inline the code and move the tests.

slaskawi: Updated, thanks!

pkg/user/apiserver/registry/user/etcd/etcd_test.go
slaskawi: Updated, thanks!

slaskawi:
@s-urbaniak @stlaz Updated - ready for the next round.

pkg/oauth/apis/oauth/validation/validation_test.go
```diff
@@ -746,3 +746,38 @@ func TestValidateAuthorizeTokensUpdate(t *testing.T) {
...
+		T    field.ErrorType
+		F    string
+	}{
+		"system:admin": {
+			T: field.ErrorTypeInvalid,
+			F: "metadata.name",
+		},
+	}
+	for k, v := range errorCases {
+		errs := ValidateUserNameField(k, fieldArg)
```
s-urbaniak, line 769: please add subtests as dave explained in the aforementioned blog post. something along the lines:
```
for name, tc := range cases {
  t.Run(name, funct(t *testing.T) {
  })
}
```

`tc` is usually used as the abbreviation for test-case.

pkg/oauth/apis/oauth/validation/validation_test.go
```diff
@@ -746,3 +746,38 @@ func TestValidateAuthorizeTokensUpdate(t *testing.T) {
...
+	valid := "kube:admin"
+	errs := ValidateUserNameField(valid, fieldArg)
+	if len(errs) != 0 {
+		t.Errorf("expected success: %v", errs)
+	}
+
+	errorCases := map[string]struct {
+		name string
+		T    field.ErrorType
+		F    string
```
s-urbaniak, line 761: don't hesitate to make this more verbose, i.e. `expected_err_type` and `expected_err_field`.

pkg/oauth/apis/oauth/validation/validation_test.go
```diff
@@ -746,3 +746,38 @@ func TestValidateAuthorizeTokensUpdate(t *testing.T) {
...
+	}{
+		"system:admin": {
+			T: field.ErrorTypeInvalid,
+			F: "metadata.name",
+		},
+	}
+	for k, v := range errorCases {
+		errs := ValidateUserNameField(k, fieldArg)
+		if len(errs) == 0 {
+			t.Errorf("expected failure %s for %v", k, k)
```
s-urbaniak, line 771: nit: more idiomatic would be to just say `t.Errorf("expected error, got none")`. If you `t.Run` the context will be clear in the output.

pkg/oauth/apis/oauth/validation/validation_test.go
```diff
@@ -746,3 +746,38 @@ func TestValidateAuthorizeTokensUpdate(t *testing.T) {
...
+			continue
+		}
+		for i := range errs {
```
s-urbaniak, line 774: is `len(errs) > 1` actually expected?

pkg/oauth/apis/oauth/validation/validation_test.go
```diff
@@ -746,3 +746,38 @@ func TestValidateAuthorizeTokensUpdate(t *testing.T) {
...
+			if errs[i].Type != v.T {
+				t.Errorf("%s: expected errors to have type %s: %v", k, v.T, errs[i])
```
s-urbaniak, line 776: suggestion: `t.Errorf("expected error type %s, got %s", tc.expected_err_type, errs[i].Type)`

pkg/oauth/apis/oauth/validation/validation_test.go
```diff
@@ -746,3 +746,38 @@ func TestValidateAuthorizeTokensUpdate(t *testing.T) {
...
+			}
+			if errs[i].Field != v.F {
+				t.Errorf("%s: expected errors to have field %s: %v", k, v.F, errs[i])
```
s-urbaniak, line 779: suggestion: `t.Errorf("expected error field %s, got %s", tc.expected_err_field, errs[i].Field)`

pkg/oauth/apis/oauth/validation/validation_test.go
s-urbaniak: if we literally expect just one error: `t.Error("expected 1 error, got %d", len(errs))`

pkg/oauth/apis/oauth/validation/validation_test.go
slaskawi: Done

pkg/oauth/apis/oauth/validation/validation_test.go
slaskawi: Done

pkg/oauth/apis/oauth/validation/validation_test.go
slaskawi: Done

pkg/oauth/apis/oauth/validation/validation_test.go
slaskawi: The `expected_err_type` returns an `ErrorList` so I could either implement it like this or add an assert ensuring there's one error. Either way - the end result is the same.

pkg/oauth/apis/oauth/validation/validation_test.go
slaskawi: I think leaving it like this is a bit more descriptive - the prefix (`"%s:[...]"`) naturally guides you to a failed use case.

pkg/oauth/apis/oauth/validation/validation_test.go
slaskawi: The same as the above.

slaskawi:
Comments addressed. Ready for the next round @s-urbaniak @stlaz

pkg/user/apiserver/registry/user/etcd/etcd.go
```diff
@@ -3,8 +3,7 @@ package etcd
 import (
 	"context"
 	"errors"
-	"strings"
-
+	"github.com/openshift/oauth-apiserver/pkg/oauth/apis/oauth/validation"
```
stlaz, line 6: move among the other openshift imports.

My preferred import order is:
system
k8s
openshift
local

stlaz (changes requested):

pkg/oauth/apis/oauth/validation/validation_test.go
```diff
@@ -746,3 +746,38 @@ func TestValidateAuthorizeTokensUpdate(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateUserNameField(t *testing.T) {
+	fieldArg := field.NewPath("metadata", "name")
+	valid := "kube:admin"
```
stlaz, line 752: nit: `validUsername`. Be verbose with variable names

pkg/oauth/apis/oauth/validation/validation_test.go
```diff
@@ -746,3 +746,38 @@ func TestValidateAuthorizeTokensUpdate(t *testing.T) {
...
+	errs := ValidateUserNameField(valid, fieldArg)
+	if len(errs) != 0 {
+		t.Errorf("expected success: %v", errs)
+	}
```
stlaz, line 756: why is this not a part of the `for` below?

stlaz (changes requested):

pkg/oauth/apis/oauth/validation/validation_test.go
```diff
@@ -746,3 +746,38 @@ func TestValidateAuthorizeTokensUpdate(t *testing.T) {
...
+			expectedErrType:  field.ErrorTypeInvalid,
+			expectedErrField: "metadata.name",
+		},
+	}
+	for name, tc := range errorCases {
+		errs := ValidateUserNameField(name, fieldArg)
+		if len(errs) == 0 {
+			t.Errorf("%s: expected 1 error got 0", name)
+			continue
+		}
```
stlaz, line 773: This is more of a legacy way of how unit tests used to be written (and I see it all over this repo, so I understand).

The more current way is more similar to what you can see here:
https://github.com/enj/cluster-authentication-operator/blob/1ca2f4502798fe963b76584678d8fc341d91b2b2/pkg/controllers/configobservation/oauth/observe_templates_test.go#L19

Notice the
```golang
tests := []struct {
		name           
```
- so no map, you already have the name be the part of the struct + you are able to determine the order in which the tests will run

and
```golang
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			indexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
```
which I believe is something @s-urbaniak might have mentioned already.

This is the tabular style of tests that comes from https://github.com/cweill/gotests (which allows allows autogeneration of such test structures in VS Code).

pkg/user/apiserver/registry/user/etcd/etcd.go
slaskawi: Do we have a code formatting settings somewhere? Sorting import manually is probably the last thing I should do...

pkg/oauth/apis/oauth/validation/validation_test.go
slaskawi: Done

pkg/oauth/apis/oauth/validation/validation_test.go
slaskawi: This is how other tests are written. You have a positive test outside of the loop and the tests for the error cases.

pkg/oauth/apis/oauth/validation/validation_test.go
slaskawi: I regenerated the tests using the latest test templates from Goland. Should be fine now.

As a side note - is there any Goland formatter with test templates? I'm asking because without it, we'll always struggle to death with what convention to use when writing tests, sorting imports etc. It's spaces vs tabs or emacs vs vi wars ;)

pkg/oauth/apis/oauth/validation/validation_test.go
slaskawi: I refactored the tests using the latest templates from Goland.

pkg/oauth/apis/oauth/validation/validation_test.go
slaskawi: I refactored the test using the latest template from Goland. If it's not sufficient, please point me out what shall I use instead.

pkg/oauth/apis/oauth/validation/validation_test.go
slaskawi: I refactored the test using the latest template from Goland. If it's not sufficient, please point me out what shall I use instead.

slaskawi:
@stlaz @s-urbaniak Ready for the next round of reviews.

pkg/user/apiserver/registry/user/etcd/etcd.go
stlaz: My IDE is capable of doing that on its own once the main groups have been manually created. This is the same as you'd do with Python library imports, too

slaskawi: I manually sorted this out. Although, without a proper formatter this problems will be getting back over and over.

stlaz (changes requested):

pkg/oauth/apis/oauth/validation/validation_test.go
```diff
@@ -746,3 +748,27 @@ func TestValidateAuthorizeTokensUpdate(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateUserNameField(t *testing.T) {
+	fieldPath := field.NewPath("metadata", "name")
+	tests := []struct {
```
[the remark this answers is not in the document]
slaskawi: Done

pkg/oauth/apis/oauth/validation/validation_test.go
```diff
@@ -746,3 +748,27 @@ func TestValidateAuthorizeTokensUpdate(t *testing.T) {
...
+		username string
+		want     field.ErrorList
+	}{
+		{
+			username: "kube:admin",
+			want:     field.ErrorList{},
+		},
+		{
+			username: "system:admin",
```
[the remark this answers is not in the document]
slaskawi: Done

slaskawi:
@stlaz @s-urbaniak Ready for the next round of reviews.

slaskawi:
@stlaz BZ reference removed.

stlaz:
/hold cancel
thank you :slightly_smiling_face:

PR #61, opened 2021-10-20
bump kube to 1.22.2 and replace usercache with its library-go version

stlaz:
The bump of library-go for the shared usercache implementation required a newer kube version, bringing these two changes together in this PR.

/assign @s-urbaniak


s-urbaniak:
in favor of #63

PR #89, opened 2023-06-26
bump k8s to 1.27.3

atiratree:


atiratree:
not really sure how we should approach the `creationTimestamp` preconditions here: https://github.com/openshift/oauth-apiserver/pull/89/commits/7a98b50f50fc5ffc3d5cd7fb707bc53aee81a1ae . We cannot be sure what was the creation time so I am just return `Now()`. Is this okay?

pkg/cmd/oauth-apiserver/cmd.go
```diff
@@ -147,16 +154,13 @@ func (o *OAuthAPIServerOptions) NewOAuthAPIServerConfig() (*apiserver.Config, er
 		&apiserverstorage.ResourceConfig{},
 		specialDefaultResourcePrefixes,
 	)
-	if len(o.RecommendedOptions.Etcd.EncryptionProviderConfigFilepath) != 0 {
```
deads2k, line 150: do we have e2e tests demonstrating this still works?

atiratree: [encryption tests](https://github.com/openshift/cluster-authentication-operator/blob/master/test/e2e-encryption/encryption_test.go) have been run on this PR (cluster from cluster bot) and they have passed successfully

atiratree:
need to investigate `oc basics can describe an OAuth access token` failure
/hold

atiratree:
I have been incorrectly registering the singlular name in useroauthaccesstoken storage -> fixed
/hold cancel

pkg/cmd/oauth-apiserver/cmd.go
```diff
@@ -123,6 +129,12 @@ func (o *OAuthAPIServerOptions) NewOAuthAPIServerConfig() (*apiserver.Config, er
 
 	serverConfig := apiserver.NewConfig()
 	serverConfig.GenericConfig.OpenAPIConfig = openapiconfig.DefaultOpenAPIConfig()
+	serverConfig.GenericConfig.OpenAPIV3Config = openapiconfig.DefaultOpenAPIConfig()
+	serverConfig.GenericConfig.AggregatedDiscoveryGroupManager = aggregated.NewResourceManager("apis")
+	// do not to install the default OpenAPI handler in the aggregated apiserver
+	// as it will be handled by openapi aggregator (both v2 and v3)
+	// non-root apiservers must set this value to false
```
ardaguclu, line 135: Can you give an example of non-root apiserver?

ardaguclu (commented):
LGTM

pkg/cmd/oauth-apiserver/cmd.go
atiratree: these are `OAuthAPIServer` and `UserAPIServer`

pkg/cmd/oauth-apiserver/cmd.go
```diff
@@ -123,6 +129,12 @@ func (o *OAuthAPIServerOptions) NewOAuthAPIServerConfig() (*apiserver.Config, er
 
 	serverConfig := apiserver.NewConfig()
 	serverConfig.GenericConfig.OpenAPIConfig = openapiconfig.DefaultOpenAPIConfig()
+	serverConfig.GenericConfig.OpenAPIV3Config = openapiconfig.DefaultOpenAPIConfig()
+	serverConfig.GenericConfig.AggregatedDiscoveryGroupManager = aggregated.NewResourceManager("apis")
```
stlaz, line 132: shouldn't this be featuregated?

pkg/user/apiserver/registry/useridentitymapping/rest.go
```diff
@@ -320,9 +319,10 @@ func unsetIdentityUser(identity *userapi.Identity) bool {
 func mappingFor(user *userapi.User, identity *userapi.Identity) (*userinternal.UserIdentityMapping, error) {
 	return &userinternal.UserIdentityMapping{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:            identity.Name,
-			ResourceVersion: identity.ResourceVersion,
-			UID:             identity.UID,
+			Name:              identity.Name,
+			ResourceVersion:   identity.ResourceVersion,
+			UID:               identity.UID,
+			CreationTimestamp: metav1.Now(),
```
stlaz, line 325: the mapping is defined by the existence of the identity so I _think_ that the more correct mapping creation is defined by `identity.CreationTimestamp`

pkg/cmd/oauth-apiserver/cmd.go
ardaguclu: As far as I can see, this was also discussed in openshift-apiserver bump and concluded that featuregated check was done in generic-apiserver https://github.com/openshift/openshift-apiserver/pull/371#discussion_r1228083743

atiratree: now it seems to me that we don't even need it, because it is initialized later 

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/config.go#L746

will investigate

we also need to update the openshift-apiserver: https://github.com/openshift/openshift-apiserver/pull/371/commits/b746cade130cd6f3a8b33c0700613bb7b83a9a8d#diff-d9072b7a1f5c28fd0867788847343611dbdd5b5a30f2fcd858cedb6655776fb9R108

But we probably do not need feature gate for the openapi as it is already GA and locked to True, right?

pkg/user/apiserver/registry/useridentitymapping/rest.go
atiratree: ok, will update to the identity one

pkg/cmd/oauth-apiserver/cmd.go
atiratree: I just read your comment @ardaguclu. Seems this was resolved in the openshift-apiserver discussion

PR #106, opened 2024-03-06
NO-ISSUE: oauth-apiserver: allow disabling PriorityAndFairness

p0lyn0mial:
pulls https://github.com/openshift/kubernetes-apiserver/pull/55 so that apf can be disabled by passing `--enable-priority-and-fairness=false` flag. 

requires also updating the operator

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -139,7 +139,7 @@ replace (
 	k8s.io/api => k8s.io/api v0.28.3
 	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.28.3
 	k8s.io/apimachinery => k8s.io/apimachinery v0.28.3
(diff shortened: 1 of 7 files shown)


pkg/cmd/oauth-apiserver/cmd.go
```diff
@@ -139,6 +144,8 @@ func (o *OAuthAPIServerOptions) NewOAuthAPIServerConfig() (*apiserver.Config, er
 		return nil, err
 	}
 
+	// we disable PriorityAndFairness for extension servers
+	serverConfig.GenericConfig.FlowControl = nil
```
p0lyn0mial, line 148: this works but it is ugly, better if we could control it via a command line flag passed from an operator.

p0lyn0mial:
/hold 

first we need https://github.com/openshift/kubernetes-apiserver/pull/55 and we also need to patch the operator.

PR #181, opened 2026-03-18
CNTRLPLANE-2989: feat: add external-oidc subcommand to oauth-apiserver

everettraven:
## Description

This PR adds a new `oauth-apiserver external-oidc` subcommand that wraps the upstream Structured Authentication behavior so that we can start the process of migrating from configuring the feature directly on the kube-apiserver to configuring it on the oauth-apiserver as a webhook authenticator. This will allow us to extend the behavior in the near future to better support enterprise customer use cases like sourcing claims to use during identity mapping from external sources.

More information on the motivations for this change and the architectural shift, see the OpenShift Enhancement Proposal: https://github.com/openshift/enhancements/pull/1907


pkg/externaloidc/handlers/authenticate.go
```diff
@@ -0,0 +1,86 @@
...
+	authenticator authenticator.Token
+}
+
+func (a *Authenticate) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	requestedTokenReviewBytes, err := io.ReadAll(req.Body)
+	if err != nil {
+		log.Printf("error reading request body: %v\n", err)
+		rw.WriteHeader(http.StatusInternalServerError)
+		return
+	}
```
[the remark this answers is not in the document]
everettraven: Only the Kubernetes API server should ever be hitting this and the payload should be pretty small. In practice, this shouldn't be an issue.

This is part of a broader feature effort that already plans to implement guardrails to ensure that the Kubernetes API server is the only workload on the cluster that can communicate with the oauth-apiserver when operating in this mode.

pkg/externaloidc/server/server.go
```diff
@@ -0,0 +1,76 @@
...
+	tokenAuthenticator authenticator.Token
+}
+
+func (i *Instance) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&i.securePort, "secure-port", "6443", "The port on which to serve HTTPS with authentication and authorization. It cannot be switched off with 0.")
+	fs.StringVar(&i.tlsPrivateKeyFile, "tls-private-key-file", "tls.key", "The file path of the private key to use for TLS connections")
+	fs.StringVar(&i.tlsCertFile, "tls-cert-file", "tls.crt", "The file path to the certificate to use for TLS connections")
+	fs.StringVar(&i.tlsMinVersion, "tls-min-version", "", fmt.Sprintf("The minimum TLS version to use for the webhook authenticator server. Must be one of %v. If not specified, it means no opinion and a default value that is subject to change over time will be used.", crypto.ValidTLSVersions()))
+	fs.StringArrayVar(&i.tlsCipherSuites, "tls-cipher-suites", []string{}, fmt.Sprintf("The TLS cipher suites to use for serving. Valid ciphers are %v. If not specified, it means no opinion and a default value that is subject to change over time will be used.", crypto.ValidCipherSuites()))
+}
```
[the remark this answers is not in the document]
everettraven: Only the Kubernetes API server should ever be hitting this and the payload should be pretty small. In practice, this shouldn't be an issue.

This is part of a broader feature effort that already plans to implement guardrails to ensure that the Kubernetes API server is the only workload on the cluster that can communicate with the oauth-apiserver when operating in this mode.

pkg/externaloidc/handlers/authenticate.go
```diff
@@ -0,0 +1,86 @@
...
+		responseTokenReview.Status = authenticationv1.TokenReviewStatus{
+			Authenticated: true,
+			User: authenticationv1.UserInfo{
+				Username: resp.User.GetName(),
+				UID:      resp.User.GetUID(),
+				Groups:   resp.User.GetGroups(),
+				Extra:    extras,
+			},
+			Audiences: resp.Audiences,
+		}
```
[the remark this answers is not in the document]
everettraven: Won't do. This is meant to mimic the Kubernetes API server token authenticator behavior. They don't pass audiences to the authenticator either.

As per the TokenReviewSpec type documentation:
>Audiences is a list of the identifiers that the resource server presented with the token identifies as. Audience-aware token authenticators will verify that the token was intended for at least one of the audiences in this list. If no audiences are provided, the audience will default to the audience of the Kubernetes apiserver.

This authenticator will not be audience aware.

As per the TokenReviewStatus type documentation:
> Audiences are audience identifiers chosen by the authenticator that are compatible with both the TokenReview and token. An identifier is any identifier in the intersection of the TokenReviewSpec audiences and the token's audiences. A client of the TokenReview API that sets the spec.audiences field should validate that a compatible audience identifier is returned in the status.audiences field to ensure that the TokenReview server is audience aware. If a TokenReview returns an empty status.audience field where status.authenticated is "true", the token is valid against the audience of the Kubernetes API server.

This authenticator will assume that the Kubernetes API server correctly handles audience awareness from webhook token authenticators (for now).

pkg/externaloidc/authenticator/jwt/jwt.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (j *JWT) AddFlags(fs *pflag.FlagSet) {
+	j.configurator.AddFlags(fs)
+}
+
+func (j *JWT) AuthenticateToken(ctx context.Context, token string) (*authenticator.Response, bool, error) {
+	authenticator := j.configurator.TokenAuthenticator()
+	if authenticator != nil {
+		return j.configurator.TokenAuthenticator().AuthenticateToken(ctx, token)
```
liouk:
```suggestion
		return authenticator.AuthenticateToken(ctx, token)
```

pkg/externaloidc/handlers/authenticate.go
```diff
@@ -0,0 +1,86 @@
...
+	return &Authenticate{
+		authenticator: authenticator,
+	}
+}
+
+type Authenticate struct {
+	authenticator authenticator.Token
+}
+
+func (a *Authenticate) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
```
liouk, line 24: Should we validate accepted HTTP methods and return 405 when appropriate?

pkg/externaloidc/authenticator/jwt/config/configurator.go
```diff
@@ -0,0 +1,194 @@
...
+	// hash the contents of the config file to prevent unnecessary churn
+	// in the underlying authenticator.
+	if err := c.handleConfigChange(ctx); err != nil {
+		return fmt.Errorf("loading configuration: %w", err)
+	}
+
+	go filesystem.WatchUntil(ctx, time.Minute, c.configFile, func() {
+		err := c.handleConfigChange(ctx)
+		if err != nil {
+			fmt.Println("error reloading configuration", err)
```
liouk: Should we use `klog` instead?

pkg/externaloidc/authenticator/jwt/config/configurator.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}, func(err error) {
+		if err != nil {
+			fmt.Println("error watching configuration", err)
```
liouk: Should we use `klog` instead?

pkg/externaloidc/handlers/authenticate.go
```diff
@@ -0,0 +1,86 @@
+package handlers
+
+import (
+	"encoding/json"
+	"io"
+	"log"
```
liouk: Should we use `klog` instead?

pkg/externaloidc/server/server.go
```diff
@@ -0,0 +1,98 @@
...
+	"time"
+
+	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/oauth-apiserver/pkg/externaloidc/handlers"
+	"github.com/spf13/pflag"
+	"k8s.io/apiserver/pkg/authentication/authenticator"
+)
+
+const (
+	authenticatePath = "/apis/oauth.openshift.io/v1/tokenreviews"
```
liouk, line 17: Would we also need health endpoints? In its oauth operation, the apiserver gets health/ready/live endpoints from the framework, but this isn't used for the oidc subcommand.

liouk (changes requested):
I've noticed we don't have any unit tests for some of the files in this PR (configurator, jwt, server) -- not sure if you're planning on adding those later, but I think we should make sure we have sufficient coverage.

pkg/externaloidc/server/server.go
everettraven: Maybe? The only thing I could think of at the moment is something that just verifies that the server is up and running. I can add one if you think it is prudent to have them now, but I am currently going with the assumption that relying on the standard deployment health checks (like the deployment being available and all pods running) are sufficient for a first pass.

I suspect that as we continue to stabilize this for production usage prior to GA that we will have better insights as to what kind of health/readiness checks might be necessary here.

I suspect we will also want to add metrics of some kind in the future, but that seems unnecessary to focus on for an initial pass so I omitted it for now.

everettraven:
>not sure if you're planning on adding those later, but I think we should make sure we have sufficient coverage.

Ah, I forgot to move over the configurator tests. Other than that, I wasn't planning on additional tests because the rest feels like testing standard library or other imported package behavior.

pkg/externaloidc/authenticator/jwt/config/configurator.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	out := &apiserver.AuthenticationConfiguration{}
+
+	err = apiserverv1.Convert_v1_AuthenticationConfiguration_To_apiserver_AuthenticationConfiguration(config, out, nil)
+	if err != nil {
+		return nil, "", fmt.Errorf("converting external representation to internal representation: %w", err)
+	}
+
+	return out, string(configHash[:]), nil
```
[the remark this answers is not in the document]
everettraven: Doesn't actually matter.

pkg/externaloidc/server/server.go
liouk: > I suspect that as we continue to stabilize this for production usage prior to GA that we will have better insights as to what kind of health/readiness checks might be necessary here.

Agreed, we should keep this in mind for GA promotion and add anything we'll deem necessary :+1:

pkg/externaloidc/handlers/authenticate_test.go
```diff
@@ -0,0 +1,232 @@
...
+					},
+					UID: "uid",
+				},
+			},
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			req := httptest.NewRequest(http.MethodPost, "/authenticate", bytes.NewReader(tc.requestBody))
```
liouk: Ideally we'd test other HTTP methods for completeness; this seems pretty straightforward to add, but I won't block this PR on this, I'll leave it up to you.

/lgtm

everettraven: Ah, good catch - I had forgotten to come back and update these tests once I added the HTTP method check. Will update.

pkg/externaloidc/server/server.go
```diff
@@ -0,0 +1,98 @@
...
+		Addr:    fmt.Sprintf("0.0.0.0:%s", i.securePort),
+		Handler: mux,
+
+		// Match default API server values as seen in
+		// https://github.com/kubernetes/apiserver/blob/9ee59078fe09d86c6dd041c05907df0cf3fba1ad/pkg/server/secure_serving.go#L165-L173
+		ReadHeaderTimeout: 32 * time.Second,
+		IdleTimeout:       90 * time.Second,
+		MaxHeaderBytes:    1 << 20, // ~1MB
+
+		ReadTimeout: 5 * time.Second,
```
sanchezl: Shorter than `ReadHeaderTimeout` value set above, limiting that to 5s also.

pkg/externaloidc/server/server.go
```diff
@@ -0,0 +1,98 @@
...
+	if len(i.tlsCipherSuites) > 0 {
+		cipherSuites = crypto.CipherSuitesOrDie(i.tlsCipherSuites)
+	}
+
+	if len(i.tlsMinVersion) > 0 {
+		tlsMinVersion = crypto.TLSVersionOrDie(i.tlsMinVersion)
+	}
+
+	srv := &http.Server{
+		Addr:    fmt.Sprintf("0.0.0.0:%s", i.securePort),
```
sanchezl: This won't work on IPv6 only cluster.

sanchezl (changes requested):

sanchezl (approved):
/lgtm

PR #182, opened 2026-03-23
CNTRLPLANE-2260: test/e2e: migrate tokenreviews test to OTE framework

wangke19:
## Summary

Migrate the tokenreviews e2e test to run under OTE/Ginkgo via `openshift-tests`.

## Historical Background: Why `go test` Was Never Running This Test

`test/e2e/tokenreviews.go` has existed since 2021 but was **never executed by `go test`**. The reason: Go only compiles files ending in `_test.go` as test code. `tokenreviews.go` (no `_test.go` suffix) is compiled as regular package code, so `go test ./test/e2e/...` never picked it up. This was a latent bug — the `TestTokenReviews` function silently sat unused for ~5 years.

Given this, there is no meaningful `go test` baseline to preserve. Re-enabling `go test` for a 5-year-stale test that was never run would require fixing and validating a separate code path (port-forward, raw HTTP, `InsecureSkipVerify`) without any historical pass data. This PR focuses on OTE only.

## What This PR Does

- **`test/e2e/e2e.go`** (new): Ginkgo entry point with `testTokenReviews(ctx, testing.TB)`.
  Uses `oauthClient.RESTClient()` for authenticated tokenreviews calls — no port-forward,
  no raw HTTP, no TLS workarounds. Computes the token resource name via sha256 hashing
  (`tokenName` helper) to correctly distinguish the stored name from the raw token value
  sent in the `TokenReview` spec.

- **`test/e2e/helpers.go`**: Convert `NewClientConfigForTest`, `PortForwardSvc`,
  `NewResourceTrashbin`, and `ResourceTrashbin.Empty` to accept `testing.TB`.

- **`cmd/oauth-apiserver-tests-ext/dependencymagnet.go`** (new): Blank import to register
  the `test/e2e` package with the OTE binary.

- **`go.mod`**: Move `github.com/onsi/ginkgo/v2` from indirect to direct.

Note: `test/e2e/tokenreviews.go` is left as-is (original upstream content). It remains dead
code as before — modifying it would be misleading since it is never executed.

## Case Design

```
Setup
  └─ Create test user "tokenreviews-e2e-testuser"
  └─ Create OAuthClient "tokenreviews-e2e-client"
  └─ Create OAuthAccessToken with secret "tokenreviews-e2e-secret-value-for-testing"

Scenario A — Invalid token
  └─ POST TokenReview with a non-existent/garbage token
  └─ Expect: authenticated = false

Scenario B — Valid token
  └─ Compute resource name: sha256~<base64(SHA256(secret))>
  └─ POST TokenReview with the real token secret
  └─ Expect: authenticated = true, username matches

Teardown
  └─ ResourceTrashbin.Empty() deletes all created resources
```

**Why SHA256?** OpenShift doesn't store raw token secrets. The `OAuthAccessToken` resource name is derived as `sha256~<base64-of-hash>`. The `tokenName()` function performs this computation so the test can reference the token by its actual resource name.

**Order matters:** Testing the invalid case first confirms the API isn't accidentally open to anything — then the valid case proves it correctly authenticates a token that does exist.

## Test Label

```
[sig-auth] OAuth should successfully review valid and invalid tokens [Component][Serial][apigroup:oauth.openshift.io]
```

Picked up by the `openshift/oauth-apiserver/component/serial` suite defined in
`cmd/oauth-apiserver-tests-ext/main.go`.

## Local Verification

```
Ran 1 of 1 Specs in 2.902 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 0 Skipped
```

## Related

- Infrastructure PR: #162 (merged ✅)
- Suite wiring + vendor PR: #163 (merged ✅)

### DIFF
--- a/cmd/oauth-apiserver-tests-ext/dependencymagnet.go
+++ b/cmd/oauth-apiserver-tests-ext/dependencymagnet.go
@@ -1,0 +1,8 @@
+// This file imports test packages to ensure they are included in the build.
+// These imports are necessary to register Ginkgo tests with the OpenShift Tests Extension framework.
+package main
+
+import (
+	// Import test packages to register Ginkgo tests
+	_ "github.com/openshift/oauth-apiserver/test/e2e"
+)
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
 	github.com/jteeuwen/go-bindata v3.0.8-0.20151023091102-a0ff2567cfb7+incompatible
+	github.com/onsi/ginkgo/v2 v2.24.0
 	github.com/openshift-eng/openshift-tests-extension v0.0.0-20250804142706-7b3ab438a292
 	github.com/openshift/api v0.0.0-20260105191300-d1c4dc4fd37b
 	github.com/openshift/apiserver-library-go v0.0.0-20251105084845-112ef95500de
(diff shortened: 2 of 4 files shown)


wangke19:
[/e2e-gcp-component-serial-disruptive-ote](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oauth-apiserver/182/pull-ci-openshift-oauth-apiserver-master-e2e-gcp-component-serial-disruptive-ote/2038505431969566720) is failed as expected, because there is no case migration for now,

PR #194, opened 2026-04-20
CNTRLPLANE-2516: Add local configuration API for external oidc mode

everettraven:
# Description

This PR adds new local copies of the upstream Kubernetes Structured Authentication Configuration APIs found at https://github.com/kubernetes/kubernetes/blob/b2f73c0d6b427e2ab5ba225375aaefc0b9bc45b2/staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1/types.go#L56 and wires up the `oauth-apiserver external-oidc` subcommand to use them for configuration of the authenticator.

The copied APIs have had anything not directly related to the JWT token authentication process removed. This is to allow us to have a baseline to begin extending from.

This PR is the first in a series of layered PRs associated with https://redhat.atlassian.net/browse/CNTRLPLANE-2516 - which targets adding an initial set of functionality to fetch identity claim information from external sources.

More details on the feature associated with this can be found in the enhancement at https://github.com/openshift/enhancements/blob/master/enhancements/authentication/external-oidc-additional-identity-information-sources.md

## Summary by CodeRabbit

* **New Features**
  * Add external OIDC JWT authentication API (internal + v1alpha1), issuer discovery, optional CA bundle, audience match policy, CEL-based claim/user validation, claim→user/group/extra mappings, and new runtime options + OIDC authenticator constructor.
* **Validation**
  * Require non-empty JWT list and add comprehensive validation (duplicates, HTTPS issuer, claim/expression rules, email verification logic).
* **Tests**
  * Add extensive unit tests covering validation scenarios.
* **Chores**
  * Include new API packages in code generation and adjust module dependency classification.

PR #195, opened 2026-04-20
CNTRLPLANE-2516: feat: externaloidc: add support for externally sourcing claims

everettraven:
# Description

This PR extends the local configuration API introduced in #194 to add support for sourcing claims from an "external" (non-token) source.

To keep the changes as small as possible, this PR only introduces the ability to source using anonymous authentication or using the token itself as an access token when making a request to the external source (useful in the scenario of wanting the same token to be able to be used for accessing an OIDC provider's [UserInfo endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo))

In order to extend the behavior to support this use case, we rely on the patch proposed in https://github.com/openshift/kubernetes-apiserver/pull/87 to introduce an extension point in the upstream OIDC authenticator behavior that we can hook into.

Most of the business logic for externally sourcing claims exists in the `pkg/externaloidc/oidc/externalclaims` package.

Other notable changes that were necessary to support this behavior is some of the custom CEL compiler, accessors, and mapper we needed to create to support the new fields that accept CEL expressions exist in the `pkg/externaloidc/cel` package.

## Summary by CodeRabbit

* **New Features**
  * External claim sources for JWT auth: fetch, map, and merge claims from external HTTPS endpoints with configurable URL/path expressions, mapping expressions, TLS, and authentication options (anonymous or request-provided token).
  * Conditional fetching using CEL expressions.

* **Validation**
  * Stricter validation and limits for external sources, mappings, conditions, hostnames, ports, and CEL expressions.

* **Tests**
  * Expanded unit and integration tests covering CEL compilation, mappers, resolver, token getters, and end-to-end OIDC flows.

* **Chores**
  * Dependency/module updates.


pkg/externaloidc/oidc/externalclaims/resolver/resolver.go
```diff
@@ -0,0 +1,354 @@
...
+		sourceMappingExpressionAccessor, ok := result.ExpressionAccessor.(*externaloidccel.ExternalSourceMappingExpression)
+		if !ok {
+			return nil, fmt.Errorf("invalid type conversion, expected ExternalSourceMappingExpression")
+		}
+
+		if result.EvalResult.Type() != cel.StringType {
+			return nil, fmt.Errorf("error evaluating external claim mapping %q: %w", sourceMappingExpressionAccessor.Claim, errors.New("expected a string return type"))
+		}
+
+		externalClaims[sourceMappingExpressionAccessor.Claim] = json.RawMessage(fmt.Sprintf("%q", result.EvalResult.Value().(string)))
```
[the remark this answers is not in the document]
everettraven: The expectation is for this to be a JSON string literal value so this is equivalent to `"\"value\""`. Because this is intended to be a string and we know the encoding format for a JSON string ahead of time we probably don't need the additional overhead of using `json.Marshal`, but I'm happy to be proven otherwise.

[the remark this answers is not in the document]
everettraven: Comparing the literal byte outputs of both approaches shows that there is no difference when comparing the byte literals that are output by each.

Because this isn't actually getting rendered out to JSON to be parsed by a JSON parser I think I'll leave it as is for now.

pkg/externaloidc/apis/authentication/types.go
```diff
@@ -0,0 +1,480 @@
...
+	// external source.
+	// When omitted or empty, claims are always attempted to be fetched
+	// from this external source.
+	// When specified, all conditions must evaluate to 'true'
+	// before claims are attempted to be fetched from this external source.
+	// conditions must not exceed 16 entries.
+	// Entries must have unique expressions.
+	// +optional
+	Conditions []ExternalSourceCondition
+}
```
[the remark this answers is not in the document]
everettraven: Fixed.

everettraven:
Blocked on https://github.com/openshift/kubernetes-apiserver/pull/87 merging.

/hold

everettraven:
Auto self approved. Removing.

/approve cancel

everettraven:
@coderabbitai review

pkg/externaloidc/apis/authentication/validation/validation.go
```diff
@@ -27,5 +34,266 @@ func ValidateAuthenticationConfiguration(compiler authenticationcel.Compiler, c
...
 
+	// Now validate our custom fields
+	for i, jwt := range c.JWT {
+		errors = append(errors, validateExternalClaimsSources(compiler, jwt.ExternalClaimsSources, root.Index(i).Child("externalClaimsSources"))...)
+	}
+
 	return errors
 }
+
+const maxExternalClaimSources = 5
```
ehearne-redhat, line 51: Do you think it is helpful to add a comment to explain why we chose 5 in particular? I know we discussed this together but for those who don't have that context might wonder. :)

pkg/externaloidc/apis/authentication/validation/validation.go
```diff
@@ -27,5 +34,266 @@ func ValidateAuthenticationConfiguration(compiler authenticationcel.Compiler, c
...
+		Expression: *condition.Expression,
+	})
+	if err != nil {
+		return field.ErrorList{field.Invalid(path.Child("expression"), *condition.Expression, fmt.Sprintf("error compiling expression: %v", err))}
+	}
+
+	return nil
+}
+
+const maxSourcedClaimMappings = 16
```
ehearne-redhat, line 207: same idea as https://github.com/openshift/oauth-apiserver/pull/195/changes#r3173536821

pkg/externaloidc/apis/authentication/validation/validation.go
```diff
@@ -27,5 +34,266 @@ func ValidateAuthenticationConfiguration(compiler authenticationcel.Compiler, c
...
+		PathExpression: *pathExpression,
+	})
+	if err != nil {
+		return field.ErrorList{field.Invalid(path, *pathExpression, fmt.Sprintf("error compiling expression: %v", err))}
+	}
+
+	return nil
+}
+
+const maxExternalSourceConditions = 16
```
ehearne-redhat, line 160: same idea as https://github.com/openshift/oauth-apiserver/pull/195/changes#r3173536821

pkg/externaloidc/apis/authentication/validation/validation.go
```diff
@@ -27,5 +34,266 @@ func ValidateAuthenticationConfiguration(compiler authenticationcel.Compiler, c
...
+	if err != nil {
+		return field.ErrorList{field.Invalid(path, *expression, fmt.Sprintf("error compiling expression: %v", err))}
+	}
+
+	return nil
+}
+
+var nameRegex = regexp.MustCompile("^([a-z_])+$")
+
+const maxSourceMappingNameLength = 256
```
ehearne-redhat, line 266: I actually don't remember if we discussed this one - why 256 ?

pkg/externaloidc/oidc/externalclaims/tokengetters/anonymous.go
```diff
@@ -0,0 +1,9 @@
+package tokengetters
+
+import "context"
+
+type Anonymous struct{}
+
+func (a *Anonymous) GetAccessToken(_ context.Context) (string, error) {
```
ehearne-redhat, line 7: Should we test this like how we test `requestprovided.go` ?

ehearne-redhat:
Hey @everettraven - mostly LGTM.  I have one concern that's blocking me from lgtm at https://github.com/openshift/oauth-apiserver/pull/195#discussion_r3174038565 . :)

pkg/externaloidc/oidc/externalclaims/resolver/resolver.go
```diff
@@ -0,0 +1,356 @@
...
+	return urlStr, nil
+}
+
+func getClaimsFromResponse(ctx context.Context, resp *http.Response, sourcedClaimsMapper externaloidccel.ExternalClaimsMapper) (k8soidc.ClaimsMap, error) {
+	responseBodyBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("error reading response body: %w", err)
+	}
+
+	input := map[string]any{}
```
everettraven: In some testing, I don't think we can actually guarantee that the response will always be able to marshal into a `map[string]any{}`. Instead, I think this will have to be just `any`.

pkg/externaloidc/oidc/externalclaims/resolver/resolver_test.go
```diff
@@ -0,0 +1,417 @@
...
+			case err != nil && len(tc.err) == 0:
+				t.Fatalf("received an unexpected error: %v", err)
+			case err != nil && len(tc.err) > 0 && !strings.Contains(err.Error(), tc.err):
+				t.Fatalf("error %v does not contain expected substring %q", err, tc.err)
+			}
+		})
+	}
+}
+
+func TestExpandClaims(t *testing.T) {
```
everettraven, line 154: As part of testing I found that we won't always be able to get outputs from external sources that unmarshal cleanly into a `map[string]any` - I'll need to update the test structure and add some more tests that exercise these more complex cases.

pkg/externaloidc/apis/authentication/validation/validation.go
everettraven: Added a comment to provide some more explanation here.

pkg/externaloidc/apis/authentication/validation/validation.go
everettraven: Added a comment to explain in more detail the reasoning for a maximum of 16 conditions.

pkg/externaloidc/apis/authentication/validation/validation.go
everettraven: Done.

pkg/externaloidc/apis/authentication/validation/validation.go
everettraven: Explained in detail the reasoning for 256 as the maximum length for an externally sourced claim name.

pkg/externaloidc/oidc/externalclaims/tokengetters/anonymous.go
everettraven: Traditionally, I probably wouldn't because this is a pretty straightforward bit of code, but it doesn't hurt and it was easy enough to add a single test that ensures this functions as expected and will fail if this changes.

everettraven:
@coderabbitai resume

pkg/externaloidc/cel/accessors.go
```diff
@@ -0,0 +1,79 @@
...
+
+	celgo "github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types/traits"
+	authenticationcel "k8s.io/apiserver/pkg/authentication/cel"
+)
+
+type ExternalClaimsMapper interface {
+	// EvalExternalClaim evaluates the given external claim and returns an EvaluationResult.
+	// This is used for external claim source validation that contains a single external claim.
+	EvalExternalClaim(context.Context, traits.Mapper) (authenticationcel.EvaluationResult, error)
```
liouk: Do we have a particular need for the singular form of this? As far as I can see this isn't actually used.

pkg/externaloidc/apis/authentication/validation/validation.go
```diff
@@ -27,5 +36,293 @@ func ValidateAuthenticationConfiguration(compiler authenticationcel.Compiler, c
...
+	if seenExternalClaimNames.Has(*name) {
+		return field.ErrorList{field.Duplicate(path, *name)}
+	}
+
+	seenExternalClaimNames.Insert(*name)
+
+	return nil
+}
+
+func validateExternalClaimsSourceTLS(compiler oidc.Compiler, tls *authentication.TLS, path *field.Path) field.ErrorList {
```
liouk: `compiler` is unused, maybe leftover?

pkg/externaloidc/oidc/externalclaims/resolver/resolver.go
```diff
@@ -0,0 +1,372 @@
...
+		klog.Errorf("external claims resolver: performing external claims request: %v", err)
+		return nil
+	}
+
+	if resp == nil {
+		return nil
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		resp.Body.Close()
```
liouk: Why not `defer` this so that it doesn't leak in case of panics?

pkg/externaloidc/apis/authentication/validation/validation.go
```diff
@@ -27,5 +36,293 @@ func ValidateAuthenticationConfiguration(compiler authenticationcel.Compiler, c
...
+	allErrs = append(allErrs, ValidateExternalClaimsSourceURLHostname(sourceURL.Hostname, path.Child("hostname"))...)
+	allErrs = append(allErrs, ValidateExternalClaimsSourceURLPathExpression(compiler, sourceURL.PathExpression, path.Child("pathExpression"))...)
+
+	return allErrs
+}
+
+const (
+	dns1123LabelFmt     string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?"
+	dns1123SubdomainFmt string = dns1123LabelFmt + "(\\." + dns1123LabelFmt + ")*"
+	optionalPortFmt     string = "(:(\\d{1,5}))?"
```
liouk: Disclaimer: used Claude to review regexes.

We could tighten the regex to reject leading zeros and exclude port 0: `(:([1-9]\d{0,4}))?`

If we want to explicitly allow port 0, then it becomes: `(:(0|[1-9]\d{0,4}))?`.

pkg/externaloidc/oidc/externalclaims/resolver/resolver_test.go
```diff
@@ -0,0 +1,493 @@
...
+				t.Fatalf("starting claims (-) after expansion does not match expected claims (+): %s", diff)
+			}
+		})
+	}
+}
+
+func newTestServer(response any, validToken string) *httptest.Server {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/claims", func(w http.ResponseWriter, r *http.Request) {
+		if r.Header.Get("Authorization") != fmt.Sprintf("Bearer %s", validToken) {
```
liouk: In case of anonymous auth, we send no Authorization header, hence this will result in 401; should we first check whether a token exists, and then the header?

Also, we can then add a happy-path test for the anonymous auth which is now missing AFAICS.

pkg/externaloidc/oidc/externalclaims/resolver/resolver.go
```diff
@@ -0,0 +1,372 @@
...
+		return "", fmt.Errorf("oidc: error getting url hostname: invalid type conversion, expected ExternalSourceURLExpression")
+	}
+
+	urlStr := fmt.Sprintf("https://%s/%s", urlExpressionAccessor.Hostname, path)
+
+	return urlStr, nil
+}
+
+func getClaimsFromResponse(ctx context.Context, resp *http.Response, sourcedClaimsMapper externaloidccel.ExternalClaimsMapper) (k8soidc.ClaimsMap, error) {
+	responseBodyBytes, err := io.ReadAll(resp.Body)
```
liouk, line 355: Claude suggests to use `io.LimitReader` instead (TIL), to avoid having a misbehaving external source OOM the apiserver.

pkg/externaloidc/oidc/externalclaims/resolver/resolver.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	externalClaims, err := getClaimsFromResponse(ctx, resp, ecr.source.mapper.Sources)
+	resp.Body.Close()
+	if err != nil {
+		klog.Errorf("external claims resolver: getting claims from response: %v", err)
+		return nil
+	}
+
+	maps.Copy(c, externalClaims)
```
liouk, line 272: IIUC `c` also contains token claims as well, and since `maps.Copy()` will overwrite claims, this means that we are giving precedence to external claims; an external source can overwrite a claim that exists on the original token.

AFAICS we are preventing duplicate claims across external sources, but not against the original token; if that's intentional, I think it would be best to clarify in a comment and in the docs.

liouk (changes requested):

pkg/externaloidc/cel/accessors.go
everettraven: Good catch, looks like I only used it in a test. I think I had a case where I used that in a past iteration that no longer is necessary. Removing.

pkg/externaloidc/oidc/externalclaims/resolver/resolver.go
everettraven: Yeah, that is intentional to keep the behavior here from becoming more complex than necessary. I'll add a comment here and update API docs to be more explicitly clear about this semantic.

>we are preventing duplicate claims across external sources

This is mostly to prevent end users from shooting themselves in the foot by overriding a claim that the resolved in an earlier source. I was debating whether or not to include this, but I wanted to start with the most aggressive validation and loosen if necessary.

pkg/externaloidc/oidc/externalclaims/resolver/resolver.go
everettraven: CodeRabbit had suggested that at some point as well with a 1MB limit.

While I understand the concern, I'm a bit hesitant to enforce that just yet because we don't really know what a reasonable limit is yet. For the use cases we are trying to solve for we don't really have any signal as to how large of a response we will receive on average.

If we feel strongly that we want to guard against excessive response sizes _now_ I'm happy to put some arbitrary number that is fairly high. Seeing as how we don't put resource limits on the deployment of the oauth-apiserver (only requests) I think the risk of us hitting OOM is fairly low at the moment.

I'm also happy add a TODO comment that links to a performance testing related Jira ticket to see if we can get a better sense of how large of a response is reasonable to expect based on feedback we've received from users re: potential no. of groups per user.

Just let me know what you'd prefer here and consider it done 👍

everettraven: Added a TODO comment for now.

liouk: Ugh, I really struggle to follow all of coderabbit's comments :/ Apologies for the repetition!

> Seeing as how we don't put resource limits on the deployment of the oauth-apiserver (only requests) I think the risk of us hitting OOM is fairly low at the moment.

Agreed -- and given that we don't have a good idea for a limit, I'm happy to revisit this once we have a way of assessing what limit to use. A jira ticket for this would be useful I think, and we could also consider adding metrics for this and revisit in the future, once we've had significant usage to review.

ehearne-redhat:
@everettraven I'm happy with the additional comments. Thanks for that! I will wait to see what you think of the AI codereview before giving lgtm.

pkg/externaloidc/oidc/externalclaims/resolver/resolver.go
everettraven: I re-used our existing testing placeholder and added a comment to it: https://redhat.atlassian.net/browse/CNTRLPLANE-2524?focusedCommentId=16928774

pkg/externaloidc/apis/authentication/validation/validation.go
```diff
@@ -27,5 +36,293 @@ func ValidateAuthenticationConfiguration(compiler authenticationcel.Compiler, c
...
+	}
+
+	_, err := compiler.CompileClaimsExpression(&cel.ExternalSourceURLExpression{
+		PathExpression: *pathExpression,
+	})
+	if err != nil {
+		return field.ErrorList{field.Invalid(path, *pathExpression, fmt.Sprintf("error compiling expression: %v", err))}
+	}
+
+	return nil
```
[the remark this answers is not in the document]
everettraven: Because the `claims` variable is made available to the path expression, and it contains arbitrary values from processing the claims in the token, we cannot guarantee at compilation time that a string list is returned. Runtime only validation is the only viable way here and silent failure is intentional to not block authentication in it's entirety if we cannot fetch claims from an external source.

Our thought here is that partial authentication > no authentication

pkg/externaloidc/oidc/externalclaims/resolver/resolver.go
```diff
@@ -0,0 +1,390 @@
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		klog.Errorf("external claims resolver: building external claims request: %v", err)
+		return nil
+	}
+
+	if accessToken != "" {
+		req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", accessToken))
+	}
```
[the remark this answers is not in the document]
everettraven: Not wiring a token getter is a programmer error and should be considered a panic.

The assumption here is that all external sources are configured with a token getter.

pkg/externaloidc/oidc/oidc.go
```diff
@@ -51,18 +58,66 @@ type Options struct {
...
-		DisallowedIssuers:    opts.DisallowedIssuers,
+		JWTAuthenticator:                     conversion.ConvertJWTAuthenticatorToApiserverJWTAuthenticator(opts.JWTAuthenticator),
+		KeySet:                               opts.KeySet,
+		CAContentProvider:                    opts.CAContentProvider,
+		Client:                               opts.Client,
+		Compiler:                             opts.Compiler,
+		SupportedSigningAlgs:                 opts.SupportedSigningAlgs,
+		DisallowedIssuers:                    opts.DisallowedIssuers,
+		ClaimsExpanders:                      externalClaimsExpanders,
+		UnknownCELValueTypesToStringListFunc: externalclaimsresolver.UnknownCELValueTypesToStringListHandler(),
```
everettraven, line 83: I need to add new unit tests that specifically exercise that this behavior is working as expected. It currently isn't and I had to open https://github.com/openshift/kubernetes-apiserver/pull/88 to fix it but having proper unit tests here would have helped catch that a bit earlier than the manual testing and verification I did and clearly forgot to test a scenario that exercised this.

everettraven:
Adding a hold until I implement those new tests.
/hold

pkg/externaloidc/cel/mappers_test.go
```diff
@@ -0,0 +1,212 @@
...
+				t.Fatalf("expected error containing %q but got none", tc.err)
+			case err != nil && len(tc.err) == 0:
+				t.Fatalf("received an unexpected error: %v", err)
+			case err != nil && len(tc.err) > 0 && !strings.Contains(err.Error(), tc.err):
+				t.Fatalf("error %v does not contain expected substring %q", err, tc.err)
+			}
+
+			if tc.resultFn != nil {
+				tc.resultFn(t, result[0].EvalResult)
+			}
```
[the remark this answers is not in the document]
everettraven: This is a test. IMO a panic here is OK if something doesn't adhere to our test expectations because it will cause the tests to fail.

everettraven:
Self-approving based on guidance from management around the new approval process for feature teams.

/approve

everettraven:
Manually tested this behavior on a clusterbot cluster and this works as expected.

/verified by @everettraven

PR #197, opened 2026-04-29
CNTRLPLANE-2518: Add support for sourcing external claims from a source that requires client credential authentication

everettraven:
# Description

This PR adds support for sourcing claims for an external source that requires an access token retrieved through the OAuth2 client credentials flow.

This is necessary for cases where:
- An ID token is used for authentication and cannot be used to access additional information on the user
- An access token used for authentication is not valid for accessing additional information on the user from this source

A prime use case for this is using an ID token and configuring the client credential flow to fetch groups from the Microsoft Graph API for scenarios where Entra ID based users belong to more than 200 groups.

## Changes

- Adds a new authentication type, `ClientCredential`, and associated `clientCredential` field that contains the configuration knobs for how the oauth-apiserver will authenticate with the external source to retrieve an access token.
- Adds a new "token-getter" implementation that fetches the access token using the OAuth2 client credentials flow, refreshing the token as needed.
- Wires in the logic for using the new token-getter when the authentication type for an external source is set to `ClientCredential

## Summary by CodeRabbit

* **New Features**
  * Fetch additional claims from external HTTP sources during authentication, with per-source TLS and auth mode.
  * OAuth2 client-credentials supported for external claim requests (initial token fetch, scopes, and optional TLS).

* **Validation**
  * Stricter validation and clearer errors for token endpoints, TLS (certificate authority now optional; at least one TLS subfield required), scopes, mappings, and conditions; limits applied.

* **Tests**
  * Expanded coverage for client-credential flows, TLS CA handling, and source validation rules.


everettraven:
Tested these changes with a clusterbot cluster and the functionality worked as expected.

/verified by @everettraven

everettraven:
@coderabbitai resume

pkg/externaloidc/oidc/externalclaims/tokengetters/clientcredential.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil
+	}
+
+	if transport == nil {
+		transport = http.DefaultTransport.(*http.Transport).Clone()
+	}
+
+	if transport.TLSClientConfig == nil {
+		transport.TLSClientConfig = &tls.Config{}
+	}
```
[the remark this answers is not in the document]
everettraven: I'm going to reject this suggestion because falling back to runtime defaults is OK for now. Go has a history of being incredibly cautious of breaking changes.

As part of adhering to platform-wide TLS configuration requirements, it is likely that we will need to provide a configuration mechanism for specifying the minimum TLS version in the API in the future.

When that is added, a new unique "modifier" function can be added to ensure the min version is set to what an end-user has configured.

pkg/externaloidc/oidc/externalclaims/tokengetters/clientcredential.go
```diff
@@ -0,0 +1,110 @@
...
+
+	ccCfg := &clientcredentials.Config{
+		ClientID:     cfg.ClientID,
+		ClientSecret: cfg.ClientSecret,
+		TokenURL:     cfg.TokenEndpoint,
+		Scopes:       cfg.Scopes,
+	}
+
+	wrappedCtx := context.WithValue(ctx, oauth2.HTTPClient, httpClient)
```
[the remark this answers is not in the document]
everettraven: I think a panic is a good signal of a programmer error, which this seems to reasonably fall under to me.

pkg/externaloidc/oidc/externalclaims/tokengetters/clientcredential.go
```diff
@@ -0,0 +1,110 @@
...
+	"time"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+
+	"github.com/openshift/oauth-apiserver/pkg/externaloidc/apis/authentication"
+)
+
+// tokenEndpointRequestTimeout is the timeout for HTTP requests to the token endpoint.
+const tokenEndpointRequestTimeout = time.Second
```
liouk, line 18: 1 sec feels a bit aggressive for a token endpoint, especially for external providers; reading the docs of this lib it also looks like there's no retry really, which is fine, but maybe we should consider increasing this slightly to cover some cases such as transient errors?

pkg/externaloidc/oidc/externalclaims/tokengetters/clientcredential.go
```diff
@@ -0,0 +1,110 @@
...
+	_, err = cc.GetAccessToken(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("getting initial access token: %v", err)
+	}
+
+	return cc, nil
+}
+
+func (cc *ClientCredential) GetAccessToken(_ context.Context) (string, error) {
+	token, err := cc.tokenSource.Token()
```
liouk, line 61: Huh, unfortunate that `Token()` doesn't accept a context for managing cancellation & deadlines 🙃

pkg/externaloidc/apis/authentication/types.go
```diff
@@ -391,28 +392,46 @@ type ExternalClaimsSource struct {
...
-	// certificateAuthority is a required field that configures the certificate authority
+	// certificateAuthority is an optional field that configures the certificate authority
 	// used to validate TLS connections with the external claims source.
 	// Must not be empty and must be a valid PEM-encoded certificate.
-	// +required
+	// +optional
 	CertificateAuthority *string
 }
 
+func (t TLS) IsZero() bool {
```
liouk: nit: The `TLS` type is always used as `*TLS`, so it'd be more idiomatic to have a pointer receiver; it would avoid a copy as well.

pkg/externaloidc/apis/authentication/validation/validation.go
```diff
@@ -324,5 +328,117 @@ func validateExternalClaimsSourceAuthentication(authn *authentication.Authentica
...
+	allErrs = append(allErrs, validateExternalClaimsSourceTLS(cfg.TLS, path.Child("tls"))...)
+
+	for i, scope := range cfg.Scopes {
+		allErrs = append(allErrs, ValidateClientCredentialConfigScope(scope, path.Child("scopes").Index(i))...)
+	}
+
+	return allErrs
+}
+
+var scopeRegex = regexp.MustCompile(`^[!#-[\]-~]+$`)
```
liouk, line 378: nit: This is quite an obscure regex 😅 Could we clarify it with a comment?

pkg/externaloidc/oidc/externalclaims/tokengetters/clientcredential.go
```diff
@@ -0,0 +1,110 @@
...
+
+	wrappedCtx := context.WithValue(ctx, oauth2.HTTPClient, httpClient)
+
+	cc := &ClientCredential{
+		tokenSource: ccCfg.TokenSource(wrappedCtx),
+	}
+
+	// do an initial access token fetch to prime
+	// the cache and catch any configuration issues.
+	_, err = cc.GetAccessToken(ctx)
```
liouk, line 52: It looks like an error here would bubble all the way up and will get the pod restarted -- is that right? Just making sure that failures here will bubble up and become visible, e.g. in pod status & events.

liouk (changes requested):

pkg/externaloidc/oidc/externalclaims/tokengetters/clientcredential.go
everettraven: Yeah, it falls back to the underlying context the token source was configured with.

I think for now that is good enough, but we may want to revisit this in the future if we are seeing any performance issues caused by access token fetching.

pkg/externaloidc/oidc/externalclaims/tokengetters/clientcredential.go
everettraven: If this errors on the initial configuration, yes.

For hot reloads, the error message will be logged an the new configuration will not take place (i.e the old configuration will still be active).

pkg/externaloidc/oidc/externalclaims/tokengetters/clientcredential.go
everettraven: I had settled on 1 second because of the fact that if we happen to hit the expiration window our external sourcing request will wait for the token request to go through.

I tried to come up with a solution that would mitigate this by refreshing in the background but it became hacky _or_ ended up with the realization that we would need to effectively write our own custom client credential behavior to keep it refreshed in the background only.

I don't mind doing that, but like with everything related to performance here I'm hesitant to over optimize until we have a more clear signal from performance testing that this is an area we need to get some performance gains from.

I'm happy to bump this a bit higher if you feel strongly.

everettraven: I'm also happy to implement a custom background refreshing cache if we feel confident we will need to do that now rather than deferring it until after we've done some initial performance testing.

pkg/externaloidc/apis/authentication/types.go
everettraven: Addressed in the latest push

pkg/externaloidc/apis/authentication/validation/validation.go
everettraven: Addressed in the latest push

pkg/externaloidc/oidc/externalclaims/tokengetters/clientcredential.go
liouk: > I had settled on 1 second because of the fact that if we happen to hit the expiration window our external sourcing request will wait for the token request to go through.

I see, that makes sense -- no strong feelings about increasing the timeout.

I agree that we should definitely first get a feeling of the performance of this and then potentially design a refresh mechanism, I don't see the immediate need to do this now for TPNU. That said, maybe we should open a GA ticket so that we make sure we decide what to do with this, even if it means we're leaving it as is.

pkg/externaloidc/oidc/externalclaims/tokengetters/clientcredential.go
liouk: That sounds reasonable. Although for hot reloads, logging an error and continuing with the old config means that the new config will be silently rejected and this will manifest only in the logs.

That's arguably better than crashing the apiserver, but what worries me is that maybe some problems don't immediately manifest while using the mechanism so it might not be immediately obvious to the admin that there's a problem with the new config.

I'm wondering whether there's a middle ground here; for example, would a status field & condition in the authentication CR make sense? This is probably beyond the scope of TPNU, but maybe worth thinking about.

pkg/externaloidc/oidc/externalclaims/tokengetters/clientcredential.go
everettraven: Jira ticket created: https://redhat.atlassian.net/browse/CNTRLPLANE-3489

pkg/externaloidc/oidc/externalclaims/tokengetters/clientcredential.go
everettraven: Yeah, it isn't the greatest experience from a troubleshooting standpoint but I think our best mitigation path here is going to end up being a best-effort combination of admission time validations (both CRD level and admission plugin level) as well as runtime validation of the generated configuration in the CAO (in which an invalid config will result in a status condition being populated).

The only thing I think we could reasonably do from this component perspective is to trigger some events, but for now at least I'd like to try and keep the scope of responsibility for this mode of operation to just being an HTTP server that does the token validation and mapping (i.e only called by the kube-apiserver and doesn't make any calls out to the kube-apiserver)

liouk: My main worry is valid configs but which fail for other reasons (e.g. endpoint problems or malfunctions) and still slip validation; but maybe we can do things like have the CAO do some kind of health/validity probe to the endpoint as well. But I think we both agree that for now this is beyond the scope of TPNU, so let's revisit this in the future :+1:

PR #212, opened 2026-07-16
CNTRLPLANE-2260: test migration of tokenreview tests to OTE

kaleemsiddiqu:
[CNTRLPLANE-2260](https://redhat.atlassian.net/browse/CNTRLPLANE-2260): tokenreview tests migrated to
OTE framework

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end coverage for the OAuth TokenReview flow, validating invalid-token rejection and successful authentication with the expected user identity.

* **Tests**
  * Integrated the OAuth e2e suite with the test extension framework.
  * Improved test helper compatibility across supported testing contexts.
  * Enhanced test cleanup to continue safely when resources are already absent and report other cleanup failures.

### DIFF
--- a/cmd/oauth-apiserver-tests-ext/main.go
+++ b/cmd/oauth-apiserver-tests-ext/main.go
@@ -61,9 +61,20 @@ func prepareOperatorTestsRegistry() (*oteextension.Registry, error) {
 	registry := oteextension.NewRegistry()
 	extension := oteextension.NewExtension("openshift", "payload", "oauth-apiserver")
 
+	// Parallel-safe, non-disruptive tests run with default (Stable) cluster
+	// health monitoring and no forced serialization.
+	extension.AddSuite(oteextension.Suite{
+		Name:    "openshift/oauth-apiserver/component/parallel",
+		Parents: []string{"openshift/conformance/parallel"},
+		Qualifiers: []string{
+			`name.contains("[Component]") && !name.contains("[Serial]") && !name.contains("[Disruptive]")`,
+		},
+	})
+
 	// Non-disruptive tests run with default (Stable) cluster health monitoring.
 	extension.AddSuite(oteextension.Suite{
 		Name:        "openshift/oauth-apiserver/component/serial",
+		Parents:     []string{"openshift/conformance/serial"},
 		Parallelism: 1,
 		Qualifiers: []string{
 			`name.contains("[Component]") && name.contains("[Serial]") && !name.contains("[Disruptive]")`,
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,8 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
 	github.com/jteeuwen/go-bindata v3.0.8-0.20151023091102-a0ff2567cfb7+incompatible
+	github.com/onsi/ginkgo/v2 v2.28.1
+	github.com/onsi/gomega v1.39.1
 	github.com/openshift-eng/openshift-tests-extension v0.0.0-20260408205138-ec501c2bf4a5
 	github.com/openshift/api v0.0.0-20260727141720-967cc4c36c9b
 	github.com/openshift/apiserver-library-go v0.0.0-20260715200723-42e5e402ca43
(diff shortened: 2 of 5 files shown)


test/e2e/e2e.go
```diff
@@ -0,0 +1,105 @@
+package e2e
```
gangwgr, line 1: is this an new file or old file? if yes so we have less changes to review

go.mod
```diff
@@ -10,6 +10,8 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
 	github.com/jteeuwen/go-bindata v3.0.8-0.20151023091102-a0ff2567cfb7+incompatible
+	github.com/onsi/ginkgo/v2 v2.28.1
```
gangwgr, line 13: why it is required? no need to update dependency

test/e2e/e2e.go
kaleemsiddiqu: This is new file.

go.mod
kaleemsiddiqu: ginko/gomega are being referred in e2e.go now, so these are required as direct dependency now.

gangwgr:
looks good to me
waiting for ci to passed

test/e2e/e2e.go
```diff
@@ -0,0 +1,105 @@
...
+
+// tokenName computes the OAuthAccessToken resource name from a raw token secret.
+// The server hashes the secret part (after "sha256~") and stores it as sha256~<base64(hash)>.
+func tokenName(rawSecret string) string {
+	h := sha256.Sum256([]byte(rawSecret))
+	return "sha256~" + base64.RawURLEncoding.EncodeToString(h[:])
+}
+
+var _ = g.Describe("[sig-auth] OAuth", func() {
+	g.It("should successfully review valid and invalid tokens [Component][Serial][apigroup:oauth.openshift.io]", func(ctx context.Context) {
```
everettraven: Does this actually need to be a serial test?

What happens if this runs in parallel with other test cases?

Is this still automatically being picked up as part of the standard conformance suites?

kaleemsiddiqu: Since test creates and deletes objects, so kept it in Serial list.
Its not part of openshift/conformance/* test suites, it is part of component test suite "openshift/oauth-apiserver/component/serial" executed through test e2e-gcp-component-serial-ote

everettraven: IIUC creating/deleting objects doesn't mean it needs to run serially.

Does anything that this test create/delete result in a component needing to rollout any new configurations?

If so, that would seem reasonable to mark the test as serial - otherwise parallel seems like it would be fine.

kaleemsiddiqu: It is not touching any cluster config. Let me move it to parallel

kaleemsiddiqu:
/testwith openshift/oauth-apiserver/master/e2e-gcp-component-parallel-ote openshift/release#84024

cmd/oauth-apiserver-tests-ext/main.go
```diff
@@ -61,6 +61,15 @@ func prepareOperatorTestsRegistry() (*oteextension.Registry, error) {
 	registry := oteextension.NewRegistry()
 	extension := oteextension.NewExtension("openshift", "payload", "oauth-apiserver")
 
+	// Parallel-safe, non-disruptive tests run with default (Stable) cluster
+	// health monitoring and no forced serialization.
+	extension.AddSuite(oteextension.Suite{
+		Name: "openshift/oauth-apiserver/component/parallel",
+		Qualifiers: []string{
+			`name.contains("[Component]") && !name.contains("[Serial]") && !name.contains("[Disruptive]")`,
```
gangwgr, line 70: instead of [Component] can we use better label

gangwgr: or keep

kaleemsiddiqu: existing serial test suites are using that, so not touching that.

cmd/oauth-apiserver-tests-ext/main.go
everettraven, line 82: Any particular reason we don't register these suites as having the respective openshift conformance suites as a parent suite?

IIUC having a parent suite would ensure that the tests within these suites would automatically be run as part of any jobs that run the associated parent suite - which seems like something we would want here.

kaleemsiddiqu: Added openshift conformance as parent.

everettraven (approved):
Changes seem fine to me.

We should ensure that the test is being picked up in jobs with this change prior to merging as part of the verification process.

/lgtm

